PR #44, opened 2022-11-28
MM: various overworld logic fixes 

Celestial-Panda:
- Entrance logic for `Ikana Castle Exterior <=> Beneath the Well end` can now expect you to use light arrows on the sun block
- Pillar chest and Goron grotto in Road to Ikana are now in logic without Epona


Nax:
Thanks a lot 👍 
Also, `GREAT_FAIRY_SWORD` was not flagged as a progression item, so the check you added had no practical effect - but I flagged it as progression right after merging this.

PR #109, opened 2022-12-08
Fix some BotW logic

XenoWars:
- Adds Kokiri Sword to Dead Hand specifically rather than the well as a whole
- Adds Din's Fire option for the map chest


data/oot/world/bottom_of_the_well.yml
```diff
@@ -12,10 +12,10 @@
     "Bottom of the Well East Cage": "has(SMALL_KEY_BOTW, 3) && has_lens"
     "Bottom of the Well Blood Chest": "has_lens"
     "Bottom of the Well Underwater 2": "can_play(SONG_ZELDA)"
-    "Bottom of the Well Map": "has_explosives || (has_bombflowers && has(SMALL_KEY_BOTW, 3))"
+    "Bottom of the Well Map": "has_explosives || (has_bombflowers && (has(SMALL_KEY_BOTW, 3) || can_use_din)"
```
Nax, line 15: Priority here is wrong, if you have din you don't need anything else

Nax:
CI was added after this, can you add an empty commit (`git commit --allow-empty`) to trigger tests?

Nax:
Thanks 👍

PR #110, opened 2022-12-08
Fix up Zora River logic

XenoWars:
- Adds a new region in Overworld.yml for Zora River Front
- Accounts for accessing Zora River as either age from all entrances
- Moves the Tree GS to Zora River Front
- Reworks logic for some of the other checks to account for both ages
- Removes ranged requirement for the Ladder GS as it can be jump slashed
- Lowered Bridge GS to Hookshot
- Adds two new macros, for diving deeper
- Adds the can_dive_big macro to Tektite grotto to cut on space

### DIFF
--- a/data/oot/macros.yml
+++ b/data/oot/macros.yml
@@ -37,3 +37,5 @@
 "has_tunic_zora": "is_adult && has(TUNIC_ZORA)"
 "has_iron_boots": "is_adult && has(BOOTS_IRON)"
 "has_hover_boots": "is_adult && has(BOOTS_HOVER)"
+"can_dive_small": "has(SCALE) || has_iron_boots"
+"can_div_big": "has(SCALE, 2) || has_iron_boots"
--- a/data/oot/macros.yml
+++ b/data/oot/macros.yml
@@ -38,4 +38,4 @@
 "has_iron_boots": "is_adult && has(BOOTS_IRON)"
 "has_hover_boots": "is_adult && has(BOOTS_HOVER)"
 "can_dive_small": "has(SCALE) || has_iron_boots"
-"can_div_big": "has(SCALE, 2) || has_iron_boots"
+"can_dive_big": "has(SCALE, 2) || has_iron_boots"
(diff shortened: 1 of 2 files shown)


data/oot/macros.yml
```diff
@@ -37,3 +37,5 @@
 "has_tunic_zora": "is_adult && has(TUNIC_ZORA)"
 "has_iron_boots": "is_adult && has(BOOTS_IRON)"
 "has_hover_boots": "is_adult && has(BOOTS_HOVER)"
+"can_dive_small": "has(SCALE) || has_iron_boots"
+"can_div_big": "has(SCALE, 2) || has_iron_boots"
```
Nax, line 41: Typo here

XenoWars: oops

data/oot/world/overworld.yml
```diff
@@ -289,21 +290,27 @@
...
+"Zora River":
+  exits:
+    "Zora River Front": "true"
     "Zora Domain": "can_play(SONG_ZELDA)"
+    "Lost Woods": "can_dive_small"
   locations:
     "Zora River Bean Seller": "is_child"
-    "Zora River HP Pillar": "true"
-    "Zora River HP Platform": "true"
+    "Zora River HP Pillar": "is_child || has_hover_boots"
```
Celestial-Panda, line 306: This (the pillar) can also be reached with beans according to main OOTR logic

data/oot/world/overworld.yml
```diff
@@ -289,21 +290,27 @@
...
   exits:
     "Hyrule Field": "true"
+    "Zora River": "is_adult || (is_child && has_explosives)"
+  locations:
+    "Zora River GS Tree": "is_child"
+"Zora River":
+  exits:
+    "Zora River Front": "true"
     "Zora Domain": "can_play(SONG_ZELDA)"
+    "Lost Woods": "can_dive_small"
```
Celestial-Panda, line 303: The way back should also be added in Lost Woods

Nax: It is I think, line 155

PR #133, opened 2022-12-13
Update OoT overworld logic for OoTR compliance

AmazingAmpharos:
This will make all of the overworld checks, if I made no errors, have the underlying same logic as they have in OoTR. I also added several convenience macros and updated some of the various logics to use them in the course of this.


Nax:
👍 Thanks

PR #135, opened 2022-12-14
OoTR logic compliance part 2

AmazingAmpharos:
This should bring Dodongo's Cavern, Jabu Jabu's Belly, Gerudo Fortress, and GTG into logical harmony with OoTR. A few minor logic bugs that could have possibly made broken seeds are fixed too (but all were pretty low probability issues), mostly it's making a few things less restrictive. This should also be construed as to include Deku Tree, Bottom of the Well, and Ice Cavern, but those three dungeons were perfect as-is so there is no file being changed.


Nax:
👍 Thanks a lot!

PR #137, opened 2022-12-14
Improve logic for Clock Town, Termina Field, and Southern Swamp

XenoWars:
- Moves Postman Hat check to East Clock Town, adding an event in Post Office for it
- Adds Deku + Moon's Tear for the lower chest in SCT
- Adds an event for meeting Anju and adds it to the Letter to Kafei and Couple's Mask checks
- Adds Deku Mask for defeating the Peahat
- Accounts for having a ranged weapon to defeat the Bad Bats in Road to Southern Swamp
- Splits the swamp into three regions: Swamp Front, Swamp Back, and Swamp Canopy
- - Swamp Front is everything accessible in the non-poisoned part
- - Swamp Back is everything accessible in the poisoned part
- - Swamp Canopy is the platforming above the poison water to get to Woodfall
- Adds event for the boat ride accounting for either the pictobox or saving Koume
- Adds event for buying beans from the Swamp Deku Scrub if you do not have access to Deku Palace


data/mm/macros.yml
```diff
@@ -14,7 +13,7 @@
 "can_use_elegy2": "can_play(SONG_EMPTINESS) && (has(MASK_ZORA) || has(MASK_GORON))"
 "can_use_elegy3": "can_play(SONG_EMPTINESS) && has(MASK_ZORA) && has(MASK_GORON)"
 "has_bombchu": "has(BOMB_BAG)"
-"has_beans": "event(MAGIC_BEANS)"
+"has_beans": "event(MAGIC_BEANS) || has(MAGIC_BEAN)"
```
Nax, line 16: Not correct - you might have found the single bean but be unsable to buy more (because no deku mask)

data/mm/world/overworld.yml
```diff
@@ -218,23 +224,23 @@
...
 "Road to Southern Swamp":
   region: ROAD_TO_SWAMP
   exits:
-    "Swamp": "true"
+    "Swamp Front": "true"
     "Termina Field": "true"
     "Swamp Archery": "true"
   locations:
-    "Road to Southern Swamp HP": "true"
+    "Road to Southern Swamp HP": "has_weapon_range"
```
Nax, line 243: It's possible to climb without killing the bat but I guess it should be a trick

Nax:
Can be merged once the bean issue is solved 👍

data/mm/world/overworld.yml
XenoWars: it is a trick in MM yes

data/mm/macros.yml
XenoWars: i originally set it that way, but figured just having the one bean would be enough for anything. I have just remembered the Well throws a wrench in that idea.

data/mm/world/overworld.yml
Nax: Allright then 👍

data/mm/macros.yml
Nax: Well + have to assume worst case usage as SoT resets beans to zero and you can also plant the wrong one

PR #143, opened 2022-12-15
OoTR Compliance Update: Forest Temple

AmazingAmpharos:
This sorts out some Forest logic, making it generally more robust. Changes include accounting for using the well as a connector, properly handling shenanigans with the ledges involving Longshot and Hover Boots, and recognizing that Din's Fire can melt the frozen switch late in the dungeon (fun fact: frozen switches only care if they thaw, not if they actually get shot).


Nax:
👍 
The GS using the event could be listed twice instead as we talked about, but it's very minor, let's merge

PR #144, opened 2022-12-15
OoTR Compliance Update: Fire Temple

AmazingAmpharos:
This updates Fire's logic to match OoTR's, insofar as non-keysanity is concerned, and it required some rather hefty refactoring. Deep fire was honestly just very wrong, and that was the main issue. Several other issues, including a very risky missing explosives check in early fire, are addressed as well.


Nax:
👍 Looks good

PR #145, opened 2022-12-15
OoTR Compliance Update: Water Temple

AmazingAmpharos:
This updates Water Temple's logic, though it required some significant creativity. I honestly cannot 100% promise this is an exact match for OoTR, but I am reasonably sure it is logically sound. The key point is that water level changing and whether you can guarantee access to any given water level is just a bit picky, and some of the logic requirements (especially for water level high) are kinda weird and don't have the most obvious tie-in to how you would physically navigate the dungeon. I also fixed some issues like accounting for Dragon Spiral access via the river and fixing the requirements in the Boss Key loop in general.

### DIFF
--- a/data/oot/world/water_temple.yml
+++ b/data/oot/world/water_temple.yml
@@ -7,26 +7,29 @@
   dungeon: Water
   exits:
     "Water Temple Ruto Room": "has_iron_boots"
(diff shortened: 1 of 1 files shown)


Nax:
👍

PR #146, opened 2022-12-15
OoTR Compliance Update: Shadow Temple

AmazingAmpharos:
This updates Shadow Temple's logic to match OoTR's. This was by far the simplest of the adult dungeons as expected, mostly because Shadow Temple is just a straight line. I did fix up a lot of Lens of Truth requirements (though there's still a little room to look at them even more carefully tbh, and they kinda don't matter because you get logically Lens locked at the start anyway). I also adjusted which scarecrow matters. Fun fact, the scarecrow before the boat is useless since Longshot can just reach the skull from the boat, but Hookshot can't reach the Scarecrow. This is almost certainly just an oops from the original devs. However, the other Scarecrow is very useful as it lets you reach the boss without the Bow if you have Longshot, and that really does come up (the fight isn't too bad if you're worried about that). Also, Hammer can't break the early bombable wall; that's an important fix!

### DIFF
--- a/data/oot/world/shadow_temple.yml
+++ b/data/oot/world/shadow_temple.yml
@@ -2,49 +2,49 @@
   dungeon: Shadow
   exits:
     "Graveyard Upper": "true"
-    "Shadow Temple Pit": "has(BOOTS_HOVER) || can_hookshot"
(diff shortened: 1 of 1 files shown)


Nax:
👍

PR #148, opened 2022-12-15
OoTR Compliance: Spirit Temple

AmazingAmpharos:
This time with feeling and without the error from before. This should make Spirit's logic correct, given the assumption that Bomb Bag -> Bombchus is always valid (true outside of full overworld ER) and that you can always buy sticks and nuts. This is pretty complex but not as bad as I thought it would be so there's that...


Nax:
Looks good! 👍

PR #164, opened 2022-12-21
MM: Hints

smak-acrophobic:
Updated item hints to be more in line with MMR


Nax:
This PR is set to merge `develop` into `master`, should be set to merge from your branch into `develop`

PR #166, opened 2022-12-21
MM: Improve logic for Great Bay Temple

XenoWars:
- General improvement to reaching checks and exits
- Removes or adds requirements based on MMR logic


Nax:
Looks good! 👍

PR #167, opened 2022-12-22
MM: Improve logic for Woodfall Temple

XenoWars:
- General improvement to reaching checks and exits
- Removes or adds requirements based on MMR logic
- Renames the Pre-boss room fairies' locations to better describe them


Nax:
👍

PR #169, opened 2022-12-22
MM: Improve logic for Snowhead Temple

XenoWars:
- General improvement of logic for the dungeon
- Adds and removes requirements based on MMR logic
- Splits some rooms into different areas to account for not having the ability to access certain parts of them
- Added macro for long distance hot spring water


Nax:
👍

PR #171, opened 2022-12-23
[OOT] 4 new sometimes hints

Celestial-Panda:
### Done
- Water Temple Central Pillar chest
- Water Temple BK chest
- Shadow Temple Flaming Pot

### Still need work
- Water Temple River chest + GS if tokensanity is on


src/common/text.c
```diff
@@ -359,7 +359,7 @@ const RegionName kRegionNamesOot[] = {
     { "in",         "the " TEXT_COLOR_YELLOW "Sacred Realm" },
     { "inside",     "the " TEXT_COLOR_GREEN "Deku Tree" },
     { "inside",     TEXT_COLOR_RED "Dodongo's Cavern" },
-    { "inside",     TEXT_COLOR_BLUE "Jabu-Jabu" },
+    { "inside",     TEXT_COLOR_BLUE "Jabu Jabu's Belly" },
```
Nax, line 362: The game has the hyphen there

src/common/text.c
```diff
@@ -368,20 +368,20 @@ const RegionName kRegionNamesOot[] = {
...
+    { "in",         TEXT_COLOR_YELLOW "Hyrule Castle" },
     { "in",         TEXT_COLOR_RED "Ganon's Castle Exterior" },
     { "in",         "the " TEXT_COLOR_GREEN "Lost Woods" },
-    { "in",         "the " TEXT_COLOR_GREEN "Sacred Meadow" },
+    { "in",         "the " TEXT_COLOR_GREEN "Sacred Forest Meadow" },
     { "in",         TEXT_COLOR_RED "Kakariko" },
-    { "inside",     TEXT_COLOR_PINK "Kakariko's Graveyard" },
-    { "on",         TEXT_COLOR_RED "Death Mountain's Trail" },
-    { "in",         TEXT_COLOR_RED "Death Mountain's Crater" },
+    { "inside",     TEXT_COLOR_PINK "Graveyard" },
```
Nax, line 382: There are two graveyards (with Ikana), so it's better to name both

src/common/text.c
Celestial-Panda: Wouldn't be on par with OOTR hints, but I suppose - [main OOTR hints code](https://github.com/TestRunnerSRL/OoT-Randomizer/blob/Dev/Hints.py#L330)

src/common/text.c
Celestial-Panda: I suppose that's fair, although MM's Graveyard is already demarcated by "Ikana" in front of it. But I'll just revert that back, not a big deal, just weird

Celestial-Panda:
Alright, now that I'm on winter break, is there a way with the current hint system to have Conditional Hints? I.e if OOT tokensanity is either set to all or dungeons, the river hint has both the skulltula and the chest, but if tokensanity is off, only the chest gets hinted?

PR #173, opened 2022-12-25
Address missing check for Captain's Hat 

AmazingAmpharos:
Oops, I realized I made this mistake a few hours later. The Oceanside Spider House chest logically requires Captain's Hat, and therefore that location needed to be in the array.


Nax:
All good 👍

PR #177, opened 2022-12-31
Logic Bug Fix

AmazingAmpharos:
My bad! This GS location didn't check for the shortcut being opened, thus allowing for impossible seeds in which Slingshot isn't available but Boomerang is.


Nax:
Happens. Thanks for the fix 👍

PR #189, opened 2023-01-12
Gerudo Fortress age logic fix

AmazingAmpharos:
The logic could expect, under a very particular set of circumstances, for child to be able to access GTG. This is problematic, and this fixes that situation.


Nax:
👍

PR #193, opened 2023-01-15
Added ability to toggle dpad items to a different set

adaliaramon:
I absolutely understand if you are not interested in this. I create the pull-request just in case, since I am using it myself.

### DIFF
--- a/src/common/dpad.c
+++ b/src/common/dpad.c
@@ -17,6 +17,8 @@ static const int kDpadOffY[] = { 1, -1, 0, 0 };
 
 static float kDpadItemScale = 0.4f;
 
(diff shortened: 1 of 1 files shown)


Nax:
It's pretty good, I just think the L button is not a very good fit for this (since it also toggles the map).

Maybe we can find a better button combo for this?

adaliaramon:
I agree this is a problem. I thought about using dpad up but it is already used in MM.

Nax:
Closing as there is no obvious way to integrate that properly (without messing with the minimap).  
I'l re-open if a discussion emerges where a clear mapping is found.

PR #196, opened 2023-01-17
Preliminary Logic Trick Implementation

AmazingAmpharos:
This implements a subset of logic tricks from OoTR and MMR that I believe are the most relevant and high impact. This could probably use a looking over, but I can confirm at least the logic end appears to work well.


Nax:
Merging on a branch right now because I want to add specific support for tricks instead of using regular settings.

PR #206, opened 2023-01-27
OOT: Fix Death Mountain Bean event

XenoWars:
Adds usage of bomb flowers to use the bean on death mountain


Nax:
I don't understand the change, you don't need bombflowers to plant the bean ?

Celestial-Panda:
You do. The soft soil is behind the rock, so you need to blow it up

XenoWars:
You can get the skulltula with a crafty boomerang trick, but planting the bean hard requires the rock destroyed

PR #238, opened 2023-02-10
Frontend names

matthewkirby:
A change to the item names written to the spoiler log. I did this via a lookup table to prevent anything about the randomizer from breaking. This has the added bonus that we can export this new function and use it to also replace the item names on the generator gui.

I know this is kind of a dirty way of doing this. If there is a different way you would prefer just let me know and I can swap it.


Nax:
Merging on a branch so I can make some edits

PR #245, opened 2023-02-15
Feature: Force junk locations

matthewkirby:
Locations can be added to `disabledLocations = []` in a config file to be disabled. A disabled check has its vanilla item added to the item pool but guarantees that it is randomized to an item from the junk pool. There is also a new section in the spoiler logs to list the disabled locations.

I had to change some existing, unrelated things in the solver to make this work. 
- I moved fixFairies, fixTokens, and fixLocations from the constructor of Solver to Solver.solve(). This means that every function that places a final item on a check is located in Solver.solve(). I had to do this because the function to disable locations needs to be after the itempool is made but before any items are placed. This way, if a location is disabled that should not be (because it is intended to be vanilla or results in impossible logic), it is caught and the randomizer will exit with an error message saying what item could not be placed.
- I changed the aforementioned functions from adding their vanilla checks to Solver.fixedLocations to simply calling Solver.place() and placing the vanilla item there. I left Solver.fixedLocations as is since I wasn't sure if it had other planned uses. I believe with this change it is now unused.

Regarding hints of disabled locations 
- Fixed Hints (e.g. MM Great Fairies): Still hinted if their checks are disabled.
- Foolish Hints: disabled locations do not contribute any weight to the foolish calculation. This means a region with every location disabled will not be hinted foolish.
- Always Hints: If the check is disabled, the always hint is not placed
- Dual Always Hints: If **both** checks are disabled, the always hint is not placed. If only one of the two checks is disabled, the dual always hint will still generate.
- Sometimes Hints: Behavior is identical to both always hints and dual always hints.
If too many hints are omitted this way, the seed will generate but crash on load. I put a super simple check in to fail to generate when this happens. Probably want to make junk hints (or something equivalent) to fill in missing spaces but that is probably beyond the scope of this PR. 

Everything I have listed up above I have extensively tested. If there is any other potential failure mode you see, please let me know and I can look into testing and fixing if need be.

Note: I have not tested this yet with the gui. My expectation (looking at the code) is that it will not impact the gui at all since it is all internal to the core repository at the moment. I wanted to first submit this PR as a yml only feature for the moment just to get review. Currently, there is no way for this to be loaded by the gui but I am going to look into making a new gui element with a simple searchbox to allow all the checks to be shown without having hundreds of lines of checkboxes.

### DIFF
--- a/lib/combo/index.ts
+++ b/lib/combo/index.ts
@@ -5,7 +5,7 @@ import { Generator } from './generator';
 import { MonitorCallbacks } from './monitor';
 import { SETTINGS, DEFAULT_SETTINGS, SETTINGS_CATEGORIES, Settings, TRICKS } from './settings';
 import { createWorld } from './logic/world';
-import { alterWorld, configFromSettings } from './logic/settings';
+import { alterWorld, configFromSettings, isShuffled } from './logic/settings';
 import { itemName } from './names';
 import { Items } from './logic/state';
 import { addItem, isDungeonItem, isDungeonReward, isJunk, isStrayFairy, isToken } from './logic/items';
@@ -17,6 +17,10 @@ type GeneratorParams = {
   opts?: OptionsInput,
   monitor?: MonitorCallbacks
 };
+
+type LocInfo = {
+  [k: string]: string[]
+}
 
 export const generate = (params: GeneratorParams): Generator => {
   const opts = options(params.opts || {});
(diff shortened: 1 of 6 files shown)


matthewkirby:
I set the PR as a draft because I want to make the locationlist accessible to the gui first. The disabled locations code is still ready for review.

matthewkirby:
Rather than writing the helper function to determine what is or isn't shuffled, I just made an empty function that always returns true. This way, when the function is fleshed out and used elsewhere in the project (a fairly large endeavor beyond the scope of this PR) it will just work. In the meanwhile, the gui will be exposed to every possible location. If a location is not shuffled but is marked as junk, the randomizer will throw an error.

PR #247, opened 2023-02-16
Add additional tricks

AmazingAmpharos:
This adds 5 new tricks for OoT and 6 new tricks for MM, bringing the total for both games to 12. Most of these tricks are simple one-line implementations that do mostly very obvious things. The only unintuitive one is the ISTT Eyegore Skips which includes the recoil flip from the updraft room, the Deku spin or jump slash down to the pre-Gomess ledge, and the Goron rolls from the Eyegore bridge down to the ledges below. This room is just so many tricks rolled into one that heavily impacts how strong players approach the dungeon so I chose this approach to encapsulate it all. The Eyegore snipe is not included due to not being compatible properly with CSMC (to my knowledge).


Nax:
Looks good!
Ideally the foolish item heuristic you added should take hints into account, but that can wait.

PR #255, opened 2023-02-19
Added a settings preset.

matthewkirby:
Export settings presets for easily sharing settings. I just made a single preset for the recently raced settings. I intended to make it the settings Ghun was planning on for future but his response when I asked what they were was "come in strim foo" 🙃. Presets can be easily updated however.

### DIFF
--- a/data/settings_presets.yml
+++ b/data/settings_presets.yml
@@ -1,5 +1,5 @@
-Standard:
-  songs: anywhere
+Standard Race:
+  songs: songLocations
   goldSkulltulaTokens: none
   housesSkulltulaTokens: none
   mapCompassShuffle: starting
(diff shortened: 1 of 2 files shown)


matthewkirby:
Converted to draft because I want to implement a test to ensure every setting is defined in each preset. The implementation does not care if a setting is not defined, it just leave it unchanged in the gui. This would result in settings not properly matching. For instance, the first pass here does not include dungeon ER (because it was not implemented when I opened the PR haha) so the preset will simply leave it unchanged. Ideally new settings can be added to any presets as they are added to the randomizer and a CI test will enforce that.

Nax:
Closing as we have a preset system now.
Consider adding various presets if you want?

PR #256, opened 2023-02-20
Added an option to generate a seed without a spoiler log.

matthewkirby:
Added a setting to disable spoiler log output. The value of the setting is appended to the random seed to make identical seeded runs different if they are run with and without a spoiler log. This is to prevent a spoiler log from being obtained by simply running the generator again with the setting swapped.

This setting is added as an advanced setting (so will appear on the gui under that heading) but it changes the cli output so a slight gui modification is required before this is merged.


matthewkirby:
Now accounted for with https://github.com/OoTMM/gui/pull/13

lib/combo/generator.ts
```diff
@@ -36,6 +36,10 @@ export class Generator {
     const buildResult = await build(this.opts);
     const rom = await pack(this.monitor, roms, buildResult, customData, this.opts);
     const { log, hash } = randomize(this.monitor, rom, this.opts);
-    return { rom, log, hash };
+    if (this.opts.settings.generateSpoilerLog) {
```
Nax, line 39: It would be better to add that check in the spoiler log pass directly (returning { log: null })

matthewkirby:
Merged dev and resolved conflicts

PR #260, opened 2023-02-21
OOT: Fix Ice Cavern logic for ER

XenoWars:
Accounts for the fact child cannot reach the map room.

### DIFF
--- a/data/oot/world/ice_cavern.yml
+++ b/data/oot/world/ice_cavern.yml
@@ -3,13 +3,13 @@
   exits:
     "Zora Fountain Frozen": "true"
   events:
-    BLUE_FIRE: "has_bottle"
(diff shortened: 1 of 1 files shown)


data/oot/world/ice_cavern.yml
```diff
@@ -3,13 +3,13 @@
   exits:
     "Zora Fountain Frozen": "true"
   events:
-    BLUE_FIRE: "has_bottle"
+    BLUE_FIRE: "has_bottle && (is_adult || event(BLUE_FIRE))"
```
Nax, line 6: It's useless to make the event require itself

Nax: Merging anyway (I'll change it quickly as I have other quickfixes to do)

PR #272, opened 2023-02-26
Update regions.ts

StunninglyWrong:
-Small tweaks to match what's shown in-game.
-Swapped two double quotes for single ones.


StunninglyWrong:
Probably one of the most useless contributions ever.
As I was skimming through core/data/regions.yml I even thought I found a missing location, because I didn't know 'Road to Mountain Village' had no checks.
At first I also wanted to add "Inside [the]" to Deku Tree and Jabu-Jabu, but it doesn't look well.

PR #297, opened 2023-03-08
Initial Fire Temple MQ Logic

AmazingAmpharos:
This was actually a lot easier than I expected.


StunninglyWrong:
I think it should say has_goron_tunic_strict

PR #298, opened 2023-03-08
Ran `npm audit fix`

adaliaramon:
Is there any reason not to do this?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2436,8 +2436,9 @@
       "peer": true
     },
     "node_modules/http-cache-semantics": {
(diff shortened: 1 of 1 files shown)


StunninglyWrong:
As a complete beginner, I asked myself and Google the same question, and I found at least one source that tells me there could be a reason: <https://overreacted.io/npm-audit-broken-by-design/>

Nax:
There is no real reason either way. None of the vulns `audit` dumps are real.

Nax:
Closing as all deps were updated

PR #324, opened 2023-03-17
Add MM Short Hookshot to logic and other logic changes

XenoWars:
- Adds all known instances of where the MM Short Hookshot can be used
- Adds some Ice Arrow logic to Woodfall Temple and Stone Tower Temple
- Adds a few more instances of Goron Bomb Jump
- Adds 8 new tricks, 4 for each game
-- OOT Child Domain Entry with Cucco
-- OOT Adult Domain Entry with Hover Boots
-- OOT Cross River of Sand with Nothing
-- OOT Drain Water Temple with Longshot
-- MM Precise Short Hookshot Usage
-- MM Backflip over Deku Palace Guards
-- MM Enter Pirate Fortress Interior with Hookshot from Boats
-- MM Inverted Stone Tower Temple Long Jump to Death Armos


XenoWars:
Reason for removing the fight reqs for the Wizzrobe is because no matter which direction you come from, you will always have something to defeat the Wizzrobe with. Getting the check you can deku void or death warp to return to the room entrance/dungeon entrance respectively.

PR #339, opened 2023-03-25
TS: Better hints in Spoiler Log

Celestial-Panda:
This PR adds 2 things:
- Better tricks, starting items, foolish regions and hints in the spoiler log
- Hash Support for offline generation. Useful to share locally generated patches and ensures you get the proper seed when patching

### DIFF
--- a/lib/cli.ts
+++ b/lib/cli.ts
@@ -42,14 +42,14 @@ const main = async () => {
     fs.readFile('./roms/mm.z64'),
   ]);
   const gen = generate({ oot, mm, opts });
-  const { rom, log, patch } = await gen.run();
+  const { rom, log, hash, patch } = await gen.run();
   await fs.mkdir('out', { recursive: true });
   const promises: Promise<unknown>[] = [];
-  promises.push(fs.writeFile('out/OoTMM.z64', rom));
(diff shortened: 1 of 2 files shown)


lib/cli.ts
```diff
@@ -42,14 +42,14 @@ const main = async () => {
...
   await fs.mkdir('out', { recursive: true });
   const promises: Promise<unknown>[] = [];
-  promises.push(fs.writeFile('out/OoTMM.z64', rom));
+  promises.push(fs.writeFile(`out/OoTMM_${hash}.z64`, rom));
   if (log)
-    promises.push(fs.writeFile('out/spoiler.txt', log));
+    promises.push(fs.writeFile(`out/spoiler_${hash}.txt`, log));
   if (patch)
-    promises.push(fs.writeFile('out/OoTMM.ootmm', patch));
+    promises.push(fs.writeFile(`out/OoTMM_${hash}.ootmm`, patch));
```
Nax, line 52: Can you:

 - Use an hyphen to separate the basename and the hash (for similarity with web)
 - Not append the hash in debug mode

Celestial-Panda: Resolved with [a6a255f](https://github.com/OoTMM/core/pull/339/commits/a6a255fe6c819b222edb9bd301ece49398ef1223)

PR #343, opened 2023-03-29
MM: Add Majora as logical time reset

XenoWars:
This adds the possibility of fighting Majora as a way to reset time instead of using song of time. I have factored in the eventual inclusion of Secret Shrine, Ikana Castle, Beneath the Well, and Pirate Fortress into the ER pool.


Nax:
Need to fix the conflict - otherwise looks good

PR #352, opened 2023-04-02
Basic overworld ER logic improvements

XenoWars:
- Adds events for MM sticks and nuts wherever they are located, and accounts for having them in logic where needed.
- Makes the SOARING region connect to every owl after having the ability to have struck the owl
- Adds trick for OOT to pass through visible one-way collisions

### DIFF
--- a/data/mm/world/overworld.yml
+++ b/data/mm/world/overworld.yml
@@ -5,7 +5,16 @@
 "SOARING":
   region: NONE
   exits:
-    "Clock Town": "has(SONG_SOARING) && event(CLOCK_TOWN)"
+    "South Clock Town": "event(CLOCK_TOWN_OWL)"
+    "Milk Road": "event(MILK_ROAD_OWL)"
+    "Swamp Front": "event(SWAMP_OWL)"
+    "Woodfall Shrine": "event(WOODFALL_OWL)"
+    "Mountain Village": "event(MOUNTAIN_VILLAGE_OWL)
+    "Snowhead Entrance": "event(SNOWHEAD_OWL)"
+    "Great Bay Coast": "event(GREAT_BAY_OWL)"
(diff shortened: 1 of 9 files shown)


XenoWars:
Will include the other ammos in this pr

data/mm/world/overworld.yml
```diff
@@ -340,6 +356,8 @@
     "Swamp Front": "true"
   events:
     KOUME: "has_red_or_blue_potion"
+    STICKS: "true"
+    "NUTS": "true"
```
Nax, line 360: No need to quote events

data/oot/world/overworld.yml
```diff
@@ -469,7 +469,7 @@
   exits:
     "Goron City": "true"
     "Dodongo Cavern": "has_bombflowers || is_adult"
-    "Kakariko": "setting(kakarikoGate, open) || event(KAKARIKO_GATE_OPEN)"
+    "Kakariko": "event(KAKARIKO_GATE_OPEN) || trick(OOT_PASS_COLLISION)"
```
Nax, line 472: This is incorrect - if you have the gate setting set to open, you don't need the event (the event wouild require that you visited kak already)

data/oot/world/overworld.yml
```diff
@@ -355,15 +355,15 @@
 "Kakariko":
   region: KAKARIKO
   events:
-    KAKARIKO_GATE_OPEN: "is_child && has(ZELDA_LETTER)"
+    KAKARIKO_GATE_OPEN: "(is_child && has(ZELDA_LETTER)) || setting(kakarikoGate, open)"
```
Nax, line 358: Better to keep it the way it was before - see comment below

data/mm/world/overworld.yml
XenoWars: Yeah this is fixed in the coming commit. i noticed it.

PR #364, opened 2023-04-06
MM: Add Support for Time-tracking in Logic

XenoWars:
This PR goes through and adds all time requirements that are needed, as accurate as necessary. Due to the nature of it, it may be a heavy pr to look through and make sure it is right. Midnight meeting still ensures deku or room key since there is no good way to account for getting kicked out. Most of the mountain region has also had its ammo drops accounted for, and there is some new regions added to account for ER in the mountains and Milk Road area.


data/mm/world/overworld.yml
```diff
@@ -897,6 +978,14 @@
     "Lone Peak Shrine Lens Chest": "true"
     "Lone Peak Shrine Boulder Chest": "has_explosives || trick_keg_explosives"
     "Lone Peak Shrine Invisible Chest": "can_use_lens"
+"Near Goron Graveyard":
+  region: MOUNTAIN_VILLAGE
+  exits:
+    "Mountain Village": "true"
+    "Goron Gravveyard": "true"
```
Celestial-Panda, line 985: Small typo here. Tests should pass once fixed

Celestial-Panda (changes requested):

Celestial-Panda (approved):

data/mm/world/overworld.yml
```diff
@@ -90,12 +89,14 @@
     "Clock Town Fairy Fountain": "true"
     "Deku Playground": "true"
   events:
-    BOMBER_CODE: "has_weapon_range"
+    BOMBER_CODE: "has_weapon_range && (first_day || second_day || final_day)"
```
Nax, line 92: Why? This will always be true

XenoWars: Because you have to do it before it passes to the next day. i wasnt sure how to make sure it can do that other than this, if it even does.

PR #474, opened 2023-08-25
MQ Pots Logic: Child and Minor Dungeons

Joshua8600:


XenoWars (approved):
Was gonna mention stuff about Jabu, but then remembered there's a switch through the water you need to press to go forward.

PR #476, opened 2023-08-27
BtW Logic Fixes and Enhancements

LumineonRL:
Fixes a few bugs with Beneath the Well's logic, as well as adding support for Dampe's Big Poe.

Summary of changes:

- Added Keese, Deku Baba, and Wallmaster souls for item events where necessary.

- Fixed a logic error for the 5 North Section pots requiring a bottle instead of a fire source.

- Fixed a logic error with the Skulltula chest not requiring a fire source.

- Added a Dampe Big Poe event and adjusted BtW logic to allow either Big Poe to be used.

- Adjusted the logic for the stick torch runs at end of the dungeon to be more robust to accommodate the Big Poe change.

- Adjusted North Section WELL_HOT_WATER, WATER, and FISH events to explicitly require a bottle even though having a bottle is currently implied by having access to the room in the first place. Both future-proofs in the event a crazy new setting revamps how BtW works and is just overall easier to read.

### DIFF
--- a/packages/core/data/mm/world/beneath_the_well.yml
+++ b/packages/core/data/mm/world/beneath_the_well.yml
@@ -7,19 +7,19 @@
 "Beneath The Well North Section":
   dungeon: BtW
   events:
-    WELL_HOT_WATER: "event(FISH) && (has_explosives || has_mask_zora || trick_keg_explosives || trick(MM_WELL_HSW))"
-    WATER: "true"
-    FISH: "true"
-    BUGS: "event(WATER) && (can_use_fire_arrows || has_sticks)"
-    BOMBS: "event(WATER) && (can_use_fire_arrows || has_sticks)"
-    RUPEES: "event(WATER) && (can_fight || has_weapon_range || has_explosives)"
+    WELL_HOT_WATER: "has_bottle && event(FISH) && (has_explosives || has_mask_zora || trick_keg_explosives || trick(MM_WELL_HSW))"
+    WATER: "has_bottle"
+    FISH: "has_bottle"
+    BUGS: "event(WATER) && has_fire_sticks"
+    BOMBS: "event(WATER) && has_fire_sticks"
+    RUPEES: "event(WATER) && (can_fight || has_weapon_range || has_explosives) && (soul_wallmaster || (event(BUGS) && soul_keese))"
   locations:
     "Beneath The Well Keese Chest": "event(WATER) && event(BUGS) && can_use_lens"
-    "Beneath The Well Pot Left Side 1": "event(WATER) && has_bottle"
-    "Beneath The Well Pot Left Side 2": "event(WATER) && has_bottle"
-    "Beneath The Well Pot Left Side 3": "event(WATER) && has_bottle"
-    "Beneath The Well Pot Left Side 4": "event(WATER) && has_bottle"
-    "Beneath The Well Pot Left Side 5": "event(WATER) && has_bottle"
+    "Beneath The Well Pot Left Side 1": "event(WATER) && has_fire_sticks"
+    "Beneath The Well Pot Left Side 2": "event(WATER) && has_fire_sticks"
+    "Beneath The Well Pot Left Side 3": "event(WATER) && has_fire_sticks"
+    "Beneath The Well Pot Left Side 4": "event(WATER) && has_fire_sticks"
+    "Beneath The Well Pot Left Side 5": "event(WATER) && has_fire_sticks"
 "Beneath The Well East Section":
   dungeon: BtW
   exits:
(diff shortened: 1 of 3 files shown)


LumineonRL:
Oh, almost forgot - also removed a duplicate check for the Skulltula chest that was in the wrong room (final instead of middle) and spelled incorrectly.

packages/core/data/mm/world/beneath_the_well.yml
```diff
@@ -7,19 +7,19 @@
 "Beneath The Well North Section":
   dungeon: BtW
   events:
-    WELL_HOT_WATER: "event(FISH) && (has_explosives || has_mask_zora || trick_keg_explosives || trick(MM_WELL_HSW))"
-    WATER: "true"
-    FISH: "true"
-    BUGS: "event(WATER) && (can_use_fire_arrows || has_sticks)"
-    BOMBS: "event(WATER) && (can_use_fire_arrows || has_sticks)"
-    RUPEES: "event(WATER) && (can_fight || has_weapon_range || has_explosives)"
+    WELL_HOT_WATER: "has_bottle && event(FISH) && (has_explosives || has_mask_zora || trick_keg_explosives || trick(MM_WELL_HSW))"
```
XenoWars, line 10: has_bottle isn't really needed here since all FISH events check for bottle regardless.

packages/core/data/mm/world/beneath_the_well.yml
```diff
@@ -29,7 +29,7 @@
   events:
     STICKS: "can_kill_baba_sticks"
     WATER: "has_bottle"
-    RUPEES: "can_fight || has_weapon_range || has_explosives"
+    RUPEES: "(can_fight || has_weapon_range || has_explosives) && soul_keese"
```
XenoWars, line 32: for what it's worth, i wouldnt consider keese as a valid source of rupees. there's a wallmaster in this section iirc, which are guaranteed rupees. Above one as well should remove Keese.

packages/core/data/mm/world/beneath_the_well.yml
```diff
@@ -67,17 +67,15 @@
     "Beneath The Well Middle Section": "true"
     "Beneath The Well Sun Block": "has(MASK_GIBDO) && soul_redead_gibdo && has_milk"
   events:
-    RUPEES: "can_fight || has_weapon_range || has_explosives"
-    BUGS: "can_use_fire_arrows || (event(WELL_BIG_POE) && has_sticks)"
-  locations:
-    "Beneath The Well Skulltulla Chest": "has(MASK_GIBDO) && soul_redead_gibdo && has_bottle"
+    RUPEES: "(can_fight || has_weapon_range || has_explosives) && soul_wallmaster"
+    BUGS: "can_use_fire_arrows || (has_sticks && has_big_poe && has(MASK_GIBDO) && soul_redead_gibdo)"
```
XenoWars, line 71: not sure why you put a Gibdo requirement here since it's trivial to just grab the bug. (should probably require bottle though)

XenoWars (changes requested):

packages/core/data/mm/world/beneath_the_well.yml
LumineonRL: The bugs are behind a web here and there's no pre-lit torch in this room.  Without Fire Arrows, you would need to light a stick in the previous room and carry it into this one, which would require giving a Big Poe to the Gibdo.

Will add bottle requirement to this.

The bugs also can't be reached without getting past a Skulltula, which either requires combat or shield, so adding that as a requirement as well.

packages/core/data/mm/world/beneath_the_well.yml
LumineonRL: Good catch - changed to Wallmaster here. Apparently Keese drop purple rupees when killed with Light Arrows so I'm adding that as a valid rupees source as well.

packages/core/data/mm/world/beneath_the_well.yml
LumineonRL: Fixed

packages/core/data/mm/world/beneath_the_well.yml
XenoWars: why shield?

LumineonRL: Shield lets you backflip past the Skulltula without killing it, which you can't do without the shield (or perhaps you can, but if you can it's non-trivial). Perhaps this would fall under the recoil flip umbrella though and we wouldn't want to include it? Will put video on Discord for reference.

LumineonRL: Logic to kill / get past the Skulltula removed. Being able to burn the web implies that you are able to get kill the Skulltula as well.

Nax:
The events should not check for a bottle, not because the bottle is required to reach them, but because the bottle is required to use them. This is checked in the various macros.

XenoWars:
> The events should not check for a bottle, not because the bottle is required to reach them, but because the bottle is required to use them. This is checked in the various macros.

Sure, if those macros actually existed. Fish, Bugs, and water do not have macros.

Nax:
All of these will have refills for scoopsanity/random bottle contents anyway, might as well future-proof by using bottle-checking macros

PR #479, opened 2023-08-30
FD "Almost" Anywhere

LumineonRL:
Adds a glitch that allows the player to transform into Fierce Deity via Sakon's Hideout on Night 3.

**Summary of changes**:

- Two new glitches for MM: one for clipping into Sakon's Hideout and another for transforming into FD from within the hideout.
- Logic for FD in the overworld, with some restrictions that are outlined below.

A description of the glitch to transform can be found here:
https://www.youtube.com/watch?v=_iSbNE8y_GU

And a description of the clip here:
https://www.youtube.com/watch?v=PQ9equzULhU

Note that accessing FD via this method is significantly more restrictive than the existing FD Anywhere setting. The mask can not be taken on and off at will - you can not put the mask back on freely once it's taken off and the glitch would need to be triggered again by returning the Hideout on night 3 before acquiring the Sun Mask.

There are a number of ways to remove the FD Mask even without being able to use it on your C-buttons. Any of the following will unequip the mask:
A) Equipping another transformation mask
B) Voiding Out
C) Dying
D) Reloading after a save+quit (including coming from OoT)

For these reasons (particularly the OoT restriction), I disabled this setting from affecting logic if Region ER and the glitch is selected but not the regular setting for the time being. I do not know what a good way to check that a route will exist to a region that will not force unequip the mask.
In addition, only logic for the overworld (minus interiors) was implemented for now.

### DIFF
--- a/packages/core/data/mm/macros.yml
+++ b/packages/core/data/mm/macros.yml
@@ -135,7 +135,7 @@ has(MASK_KAMARO) || has(MASK_GIBDO) || has(MASK_GARO)) || has(MASK_CAPTAIN)" # G
 
 # Misc
 "soul(x)": "(!setting(enemySoulsMm)) || has(x)"
-"fd_anywhere": "has(MASK_FIERCE_DEITY) && (setting(fierceDeityAnywhere) || (glitch(MM_FD_ANYWHERE) && has_regular_mask && is_night3 && !er_enabled && (can_play(SONG_SOARING) || event(SUN_MASK))))"
+"fd_anywhere": "has(MASK_FIERCE_DEITY) && (setting(fierceDeityAnywhere) || (glitch(MM_FD_ANYWHERE) && has_regular_mask && is_night3 && !setting(erRegions, none) && (can_play(SONG_SOARING) || event(SUN_MASK))))"
 "fd_goron_region_access": fd_anywhere && (has(OWL_MOUNTAIN_VILLAGE) || has(OWL_SNOWHEAD))
 "fd_zora_region_access": fd_anywhere && (has(OWL_GREAT_BAY) || has(OWL_ZORA_CAPE))
 "fd_ic_access": "fd_anywhere && (has(OWL_IKANA_CANYON) || has(OWL_STONE_TOWER))"
--- a/packages/core/data/mm/macros.yml
+++ b/packages/core/data/mm/macros.yml
@@ -124,8 +124,15 @@
 "has_mask_keaton": "has(MASK_KEATON) || has(SHARED_MASK_KEATON)"
 "has_mask_zora": "has(MASK_ZORA) || has(SHARED_MASK_ZORA)"
 "has_mask_goron": "has(MASK_GORON) || has(SHARED_MASK_GORON)"
+"has_regular_mask": "has_mask_bunny || has_mask_truth || has(MASK_BLAST || has(MASK_POSTMAN) || has(MASK_BREMEN) ||
+                    has(MASK_KAFEI) || has(MASK_TROUPE_LEADER) || has(MASK_COUPLE) || has_mask_keaton || has(MASK_KAMARO) || has(MASK_SCENTS ||
+                    has(MASK_DON_GERO) || has(MASK_BUNNY) has(MASK_STONE) || has(MASK_GARO)) || has(MASK_GIBDO) || has(MASK_GIANT)"
 
 "er_enabled": "!setting(erDungeons, none) || !setting(erIndoors, none) || !setting(erRegions, none) || !setting(erBoss, none)"
 
 # Misc
 "soul(x)": "(!setting(enemySoulsMm)) || has(x)"
+"fd_anywhere": "has(MASK_FIERCE_DEITY) && (setting(fierceDeityAnywhere) || (glitch(MM_FD_ANYWHERE) && has_regular_mask && is_night3 && !er_enabled && (can_play(SONG_SOARING) || event(SUN_MASK))))"
+"fd_goron_region_access": fd_anywhere && (has(OWL_MOUNTAIN_VILLAGE) || has(OWL_SNOWHEAD))
+"fd_zora_region_access": fd_anywhere && (has(OWL_GREAT_BAY) || has(OWL_ZORA_CAPE))
+"fd_ic_access": "fd_anywhere && (has(OWL_IKANA_CANYON) || has(OWL_STONE_TOWER))"
--- a/packages/core/data/mm/macros.yml
+++ b/packages/core/data/mm/macros.yml
@@ -124,9 +124,12 @@
 "has_mask_keaton": "has(MASK_KEATON) || has(SHARED_MASK_KEATON)"
 "has_mask_zora": "has(MASK_ZORA) || has(SHARED_MASK_ZORA)"
 "has_mask_goron": "has(MASK_GORON) || has(SHARED_MASK_GORON)"
-"has_regular_mask": "has_mask_bunny || has_mask_truth || has(MASK_BLAST || has(MASK_POSTMAN) || has(MASK_BREMEN) ||
-                    has(MASK_KAFEI) || has(MASK_TROUPE_LEADER) || has(MASK_COUPLE) || has_mask_keaton || has(MASK_KAMARO) || has(MASK_SCENTS ||
-                    has(MASK_DON_GERO) || has(MASK_BUNNY) has(MASK_STONE) || has(MASK_GARO)) || has(MASK_GIBDO) || has(MASK_GIANT)"
+
+
+"has_regular_mask": "has(MASK_POSTMAN) || has(MASK_ALL_NIGHT) || has(MASK_BLAST) || has(MASK_STONE) || has(MASK_GREAT_FAIRY) ||
+has_mask_keaton || has(MASK_BREMEN) || has_mask_bunny || has(MASK_DON_GERO) ||
+has(MASK_ROMANI) || has(MASK_TROUPE_LEADER) || has(MASK_KAFEI) || has(MASK_COUPLE) || has_mask_truth ||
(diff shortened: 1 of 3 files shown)


packages/core/data/mm/macros.yml
```diff
@@ -124,8 +124,13 @@
 "has_mask_keaton": "has(MASK_KEATON) || has(SHARED_MASK_KEATON)"
 "has_mask_zora": "has(MASK_ZORA) || has(SHARED_MASK_ZORA)"
 "has_mask_goron": "has(MASK_GORON) || has(SHARED_MASK_GORON)"
+"has_regular_mask": "has(MASK_POSTMAN) || has(MASK_ALL_NIGHT) || has(MASK_BLAST) || has(MASK_STONE) || has(MASK_GREAT_FAIRY) || has_mask_keaton || has(MASK_BREMEN) || has_mask_bunny || has(MASK_DON_GERO) || has(MASK_SCENTS) || has(MASK_ROMANI) || has(MASK_TROUPE_LEADER) || has(MASK_KAFEI) || has(MASK_COUPLE) || has_mask_truth || has(MASK_KAMARO) || has(MASK_GIBDO) || has(MASK_GARO) || has(MASK_CAPTAIN)"
 
 "er_enabled": "!setting(erDungeons, none) || !setting(erIndoors, none) || !setting(erRegions, none) || !setting(erBoss, none)"
 
 # Misc
 "soul(x)": "(!setting(enemySoulsMm)) || has(x)"
+"fd_anywhere": "has(MASK_FIERCE_DEITY) && (setting(fierceDeityAnywhere) || (glitch(MM_FD_ANYWHERE) && has_regular_mask && is_night3 && !setting(erRegions, none) && (can_play(SONG_SOARING) || event(SUN_MASK))))"
```
XenoWars, line 133: this can cause issues if you only require soaring. it can and will expect you to do things without having any owls.

packages/core/data/mm/world/overworld.yml
```diff
@@ -274,16 +274,16 @@
...
   region: CLOCK_TOWER_ROOF
   events:
     MAGIC: "can_play_time"
   exits:
     "Moon": "can_play(SONG_ORDER) && special(MOON)"
   locations:
-    "Clock Tower Roof Skull Kid Ocarina": "has_weapon_range && can_play_time"
-    "Clock Tower Roof Skull Kid Song of Time": "has_weapon_range && can_play_time"
+    "Clock Tower Roof Skull Kid Ocarina": "(has_weapon_range || (fd_anywhere && has_magic)) && can_play_time"
+    "Clock Tower Roof Skull Kid Song of Time": "(has_weapon_range || (fd_anywhere && has_magic)) && can_play_time"
```
XenoWars, line 286: unsure why magic would be needed if you could just jump attack him. surely he's tall enough where skull kid being at max height would not matter.

packages/core/data/mm/world/overworld.yml
```diff
@@ -139,8 +139,8 @@
     CLOCK_TOWN_SCRUB: "has(MOON_TEAR)"
     MAIL_LETTER: "has(LETTER_TO_KAFEI) && before(DAY2_AM_11_30)"
   locations:
-    "Clock Town South Chest Lower": "can_hookshot || (has(MASK_DEKU) && event(CLOCK_TOWN_SCRUB)) || trick(MM_SCT_NOTHING) || can_goron_bomb_jump"
-    "Clock Town South Chest Upper": "(can_hookshot || (has(MASK_DEKU) && event(CLOCK_TOWN_SCRUB)) || (can_goron_bomb_jump && can_hookshot_short)) && final_day"
+    "Clock Town South Chest Lower": "can_hookshot || (has(MASK_DEKU) && event(CLOCK_TOWN_SCRUB)) || trick(MM_SCT_NOTHING) || can_goron_bomb_jump || fd_anywhere"
+    "Clock Town South Chest Upper": "(can_hookshot || (has(MASK_DEKU) && event(CLOCK_TOWN_SCRUB)) || (can_goron_bomb_jump && can_hookshot_short) || fd_anywhere) && final_day"
```
XenoWars, line 143: This check would be considered as part of MMR's "FD Difficult Jumps" which i think we're doing parity with when FD Anywhere is properly logical.

packages/core/data/mm/world/overworld.yml
```diff
@@ -720,7 +720,7 @@
   exits:
     "Road to Southern Swamp": "true"
     "Tourist Information": "true"
-    "Swamp Back": "event(BOAT_RIDE) || event(CLEAN_SWAMP) || has_mask_zora || (has(MASK_DEKU) && (has_arrows || can_hookshot_short))"
+    "Swamp Back": "event(BOAT_RIDE) || event(CLEAN_SWAMP) || has_mask_zora || (has(MASK_DEKU) && (has_arrows || can_hookshot_short) || fd_anywhere)"
```
XenoWars, line 723: This would likely be its own trick, but i'd have to look back at MMR logic. It would be a Big Octo skip.

packages/core/data/mm/world/overworld.yml
```diff
@@ -752,23 +752,23 @@
     PICTURE_SWAMP: "has(PICTOGRAPH_BOX)"
     PICTURE_BIG_OCTO: "has(PICTOGRAPH_BOX) && soul_octorok"
   exits:
-    "Swamp Front": "event(BOAT_RIDE) || event(CLEAN_SWAMP) || has_mask_zora || ((has_arrows || can_hookshot) && (has(MASK_DEKU) || has_mask_goron))"
+    "Swamp Front": "event(BOAT_RIDE) || event(CLEAN_SWAMP) || has_mask_zora || ((has_arrows || can_hookshot) && (has(MASK_DEKU) || has_mask_goron)) || fd_anywhere"
     "Deku Palace Front": "true"
-    "Near Swamp Spider House": "has(MASK_DEKU) || has_mask_zora || event(CLEAN_SWAMP)"
-    "Swamp Canopy Back": "event(CLEAN_SWAMP)"
+    "Near Swamp Spider House": "has(MASK_DEKU) || has_mask_zora || event(CLEAN_SWAMP) || fd_anywhere"
+    "Swamp Canopy Back": "event(CLEAN_SWAMP) || fd_anywhere"
```
XenoWars, line 758: This would be FD Difficult Jumps

packages/core/data/mm/world/overworld.yml
```diff
@@ -752,23 +752,23 @@
...
 "Near Swamp Spider House":
   region: SOUTHERN_SWAMP
   exits:
     "Swamp Spider House": "has_sticks || has_arrows"
-    "Swamp Back": "has(MASK_DEKU) || has_mask_zora || event(CLEAN_SWAMP)"
-    "Near Swamp Grotto": "has(MASK_DEKU) || has_mask_zora || event(CLEAN_SWAMP)"
+    "Swamp Back": "has(MASK_DEKU) || has_mask_zora || event(CLEAN_SWAMP) || fd_anywhere"
```
XenoWars, line 763: FD Difficult Jumps

packages/core/data/mm/world/overworld.yml
```diff
@@ -909,13 +909,13 @@
   exits:
     "Near Swamp Grotto": "true"
     "Deku Palace Cliff": "true"
-    "Swamp Canopy Back": "has(MASK_DEKU)"
+    "Swamp Canopy Back": "has(MASK_DEKU) || fd_anywhere"
```
XenoWars, line 912: FD Difficult Jumps

packages/core/data/mm/world/overworld.yml
```diff
@@ -909,13 +909,13 @@
...
 "Swamp Canopy Back":
   region: SOUTHERN_SWAMP
   exits:
-    "Swamp Back": "has(MASK_DEKU) || has_mask_zora || event(CLEAN_SWAMP)"
+    "Swamp Back": "has(MASK_DEKU) || has_mask_zora || event(CLEAN_SWAMP) || fd_anywhere"
     "Woodfall": "true"
-    "Swamp Canopy Front": "has(MASK_DEKU)"
+    "Swamp Canopy Front": "has(MASK_DEKU) || fd_anywhere"
```
XenoWars, line 918: I think you get the picture.

packages/core/data/mm/macros.yml
```diff
@@ -124,8 +124,13 @@
...
+"fd_goron_region_access": fd_anywhere && (has(OWL_MOUNTAIN_VILLAGE) || has(OWL_SNOWHEAD))
+"fd_zora_region_access": fd_anywhere && (has(OWL_GREAT_BAY) || has(OWL_ZORA_CAPE))
+"fd_ic_access": "fd_anywhere && (has(OWL_IKANA_CANYON) || has(OWL_STONE_TOWER))"
```
XenoWars, line 136: These unnecessarily always force the owls if you want to use FD Anywhere even with having the setting on.

packages/core/data/mm/world/overworld.yml
```diff
@@ -1837,13 +1837,13 @@
   exits:
     "Ikana Valley": "true"
   events:
-    SUN_MASK: "(can_fight || has_explosives || has_arrows) && soul_deku_baba && soul_wolfos"
+    SUN_MASK: "(can_fight || has_explosives || has_arrows) && soul_deku_baba && soul_wolfos && is_night3"
```
XenoWars, line 1840: Inaccurate. It can be any time on day 3.

XenoWars (changes requested):
A lot of these I didn't bother looking at since it would just be listing the same things. When we properly do FD logic, we are going to have parity with MMR except it will not be locked behind turning another setting on to enable the logic.

packages/core/data/mm/macros.yml
XenoWars: Thinking about it, i think everywhere that isnt swamp, field, or clock town actually uses the following macros, so it should be fine?

PR #482, opened 2023-09-04
Glitch megaflip

CelestialKitsune:
This PR brings logic to a new glitch: megaflips
I have added all the places I thought were logically relevant, but would prefer to have a review before that goes into production.

### DIFF
--- a/packages/core/data/oot/macros.yml
+++ b/packages/core/data/oot/macros.yml
@@ -166,6 +166,7 @@
 "glitch_equip_swap": "glitch(OOT_EQUIP_SWAP) && (has(SPELL_FIRE) || (has_sticks && (is_child || setting(agelessSticks))))"
 "glitch_equip_swap_anyage": "glitch(OOT_EQUIP_SWAP) && (has(SPELL_FIRE) || (has_sticks && setting(agelessSticks)))"
 "glitch_ocarina_items": "glitch(OOT_OCARINA_ITEMS) && (has_bugs || has_fish)"
+"glitch_megaflip": "glitch(OOT_MEGAFLIP) && has_bombs && has_shield"
 
 # Misc
 "soul(x)": "(!setting(enemySoulsOot)) || has(x)"
--- a/packages/core/data/oot/macros.yml
+++ b/packages/core/data/oot/macros.yml
@@ -167,7 +167,7 @@
 "glitch_equip_swap": "glitch(OOT_EQUIP_SWAP) && (has(SPELL_FIRE) || (has_sticks && (is_child || setting(agelessSticks))))"
 "glitch_equip_swap_anyage": "glitch(OOT_EQUIP_SWAP) && (has(SPELL_FIRE) || (has_sticks && setting(agelessSticks)))"
 "glitch_ocarina_items": "glitch(OOT_OCARINA_ITEMS) && (has_bugs || has_fish)"
-"glitch_megaflip": "glitch(OOT_MEGAFLIP) && has_bombs && has_shield_for_scrubs"
+"glitch_megaflip": "glitch(OOT_MEGAFLIP) && has_bombs && has_one_hand_shield"
 
 # Misc
 "soul(x)": "(!setting(enemySoulsOot)) || has(x)"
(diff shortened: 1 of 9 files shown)


packages/core/data/oot/macros.yml
```diff
@@ -166,6 +166,7 @@
 "glitch_equip_swap": "glitch(OOT_EQUIP_SWAP) && (has(SPELL_FIRE) || (has_sticks && (is_child || setting(agelessSticks))))"
 "glitch_equip_swap_anyage": "glitch(OOT_EQUIP_SWAP) && (has(SPELL_FIRE) || (has_sticks && setting(agelessSticks)))"
 "glitch_ocarina_items": "glitch(OOT_OCARINA_ITEMS) && (has_bugs || has_fish)"
+"glitch_megaflip": "glitch(OOT_MEGAFLIP) && has_bombs && has_shield"
```
XenoWars, line 169: Should probably be a different macro since child can't use hylian shield for it.

packages/core/data/oot/world/jabu_jabu.yml
```diff
@@ -7,7 +7,7 @@
   dungeon: JJ
   exits:
     "Jabu-Jabu": "true"
-    "Jabu-Jabu Pre-Boss": "event(BIG_OCTO) || (has_hover_boots && trick(OOT_JABU_BOSS_HOVER)) || climb_anywhere"
+    "Jabu-Jabu Pre-Boss": "event(BIG_OCTO) || (has_hover_boots && trick(OOT_JABU_BOSS_HOVER)) || climb_anywhere || glitch_megaflip"
```
XenoWars, line 10: This would also require the boxless entry.

packages/core/data/oot/world/overworld.yml
```diff
@@ -1574,11 +1574,11 @@
     "Death Mountain Crater Warp": "can_hookshot || (has_hover_boots && (is_adult || has_tunic_goron_strict))"
     "Death Mountain Crater Top": "is_adult || has_tunic_goron_strict"
     "Death Mountain Crater Scrub Grotto": "can_hammer"
-    "Fairy Fountain Double Magic": "can_hammer || climb_anywhere || hookshot_anywhere"
+    "Fairy Fountain Double Magic": "can_hammer || climb_anywhere || hookshot_anywhere || glitch_megaflip"
```
XenoWars, line 1577: this would have to be from the top wouldn't it?

packages/core/data/oot/world/overworld.yml
```diff
@@ -1302,7 +1302,7 @@
   locations:
     "Zora River Bean Seller": "is_child && can_use_wallet(1)"
     "Zora River HP Pillar": "is_child || has_hover_boots || climb_anywhere || hookshot_anywhere"
```
XenoWars, line 1304: Could probably include it here as well.

XenoWars (changes requested):

packages/core/data/oot/macros.yml
```diff
@@ -166,7 +167,7 @@
 "glitch_equip_swap": "glitch(OOT_EQUIP_SWAP) && (has(SPELL_FIRE) || (has_sticks && (is_child || setting(agelessSticks))))"
 "glitch_equip_swap_anyage": "glitch(OOT_EQUIP_SWAP) && (has(SPELL_FIRE) || (has_sticks && setting(agelessSticks)))"
 "glitch_ocarina_items": "glitch(OOT_OCARINA_ITEMS) && (has_bugs || has_fish)"
-"glitch_megaflip": "glitch(OOT_MEGAFLIP) && has_bombs && has_shield"
+"glitch_megaflip": "glitch(OOT_MEGAFLIP) && has_bombs && has_shield_for_scrubs"
```
XenoWars, line 170: Surely this should be the new macro you made, right?

XenoWars (changes requested):

packages/core/data/oot/macros.yml
CelestialKitsune: No because this doesn't account for mirror shield

XenoWars (changes requested):
Just resolve the conflicts and you're good to go

PR #530, opened 2023-10-11
can_damage now includes can_use_bow

ggkfox:
the can_damage macro should include can_use_bow especially since can_use_slingshot is already included.


ggkfox:
also i should note... i double check some enemies, and they were not checking for bow. 
for example the wolfos in the sacred meadow do not check for the bow independantly of can_damage.

XenoWars:
Pretty sure bow isn't included because adult always has a sword, so it's redundant since child cannot actually use a bow.

XenoWars:
Also if you mean the wolfos that is only there for child.. it's only there as child so bow is useless. and the grotto if adult is there he can just use his sword. anything that isn't "has_weapon" is requirements for child to do damage.

PR #576, opened 2023-11-18
Weird DC stairs logic

ggkfox:
Please double check this, because there may be something that im unaware of... 
but blue fire arrows should not help you get up the stairs in DC. perhaps this logic was innitially intended for the compas room? and if there is some sort of blue fire interaction, arrows should not be here though.


XenoWars:
Might be a relic of when i thought the BFA would use the actual blue fire

ggkfox:
ah that makes sense. i tried shooting BFA all over the place there.

PR #577, opened 2023-11-21
Dc hoverboots

ggkfox:
hoverboots on child do not allow you to get to the bomb bag. the child still cannot grab the ledge bc he is too short.


XenoWars:
Yes they do

PR #585, opened 2023-11-26
DMT first spider behind wall

ggkfox:
if you had the hammer, there were missing ways to obtain the spider once the wall was broken.

### DIFF
--- a/packages/core/data/oot/world/overworld.yml
+++ b/packages/core/data/oot/world/overworld.yml
@@ -1191,7 +1191,7 @@
   locations:
     "Death Mountain Chest": "has_explosives_or_hammer"
     "Death Mountain HP": "true"
-    "Death Mountain GS Entrance": "has_explosives || (can_hammer && (has_ranged_weapon || can_use_din))"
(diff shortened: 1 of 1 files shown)


packages/core/data/oot/world/overworld.yml
```diff
@@ -1191,7 +1191,7 @@
   locations:
     "Death Mountain Chest": "has_explosives_or_hammer"
     "Death Mountain HP": "true"
-    "Death Mountain GS Entrance": "has_explosives || (can_hammer && (has_ranged_weapon || can_use_din))"
+    "Death Mountain GS Entrance": "has_explosives || (can_hammer && (is_adult || can_use_sticks || can_use_sword_goron || has_ranged_weapon_child || can_use_din))"
```
XenoWars, line 1194: instead of using the macro for goron swords, use the following:
```suggestion
    "Death Mountain GS Entrance": "has_explosives || (can_hammer && (is_adult || can_use_sticks || (age_sword_adult && has(SWORD_BIGGORON)) || has_ranged_weapon_child || can_use_din))"
```
This is because a broken giant's knife would not reach.

XenoWars (changes requested):

PR #596, opened 2023-12-20
fix MQ Fire Hammer chest being in logic for child

eedefeed:
The chest spawns on a ledge that is too high too jump up to as child. 

(I considered the possibility of spawning the chest as child then collecting as adult, but I don't think child has any options to kill the flare dancer that adult does not have access to. If there is another way, it's not already in logic so that could be added by someone later)

### DIFF
--- a/packages/data/src/world/mq/fire_temple_mq.yml
+++ b/packages/data/src/world/mq/fire_temple_mq.yml
@@ -25,7 +25,7 @@
 "Fire Temple Vanilla Hammer Loop":
   dungeon: Fire
   locations:
-    "MQ Fire Temple Hammer Chest": "has_weapon && soul_keese && soul_enemy(SOUL_ENEMY_STALFOS) && soul_iron_knuckle && soul_enemy(SOUL_ENEMY_FLARE_DANCER) && (has_bombs || can_hammer || can_hookshot)"
+    "MQ Fire Temple Hammer Chest": "(is_adult || can_hookshot || climb_anywhere) && has_weapon && soul_keese && soul_enemy(SOUL_ENEMY_STALFOS) && soul_iron_knuckle && soul_enemy(SOUL_ENEMY_FLARE_DANCER) && (has_bombs || can_hammer || can_hookshot)"
     "MQ Fire Temple Map Chest": "can_hammer && has_weapon && soul_keese && soul_enemy(SOUL_ENEMY_STALFOS) && soul_iron_knuckle && soul_enemy(SOUL_ENEMY_FLARE_DANCER)"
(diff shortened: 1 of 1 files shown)


packages/core/data/oot/world_mq/fire_temple_mq.yml
```diff
@@ -25,7 +25,7 @@
 "Fire Temple Vanilla Hammer Loop":
   dungeon: Fire
   locations:
-    "MQ Fire Temple Hammer Chest": "has_weapon && soul_keese && soul_enemy(SOUL_ENEMY_STALFOS) && soul_iron_knuckle && soul_enemy(SOUL_ENEMY_FLARE_DANCER) && (has_bombs || can_hammer || can_hookshot)"
+    "MQ Fire Temple Hammer Chest": "is_adult && has_weapon && soul_keese && soul_enemy(SOUL_ENEMY_STALFOS) && soul_iron_knuckle && soul_enemy(SOUL_ENEMY_FLARE_DANCER) && (has_bombs || can_hammer || can_hookshot)"
```
XenoWars, line 28: instead of just checking for adult, can do ageless hookshot as well. i did that for vanilla fire too. there is also climb anywhere.

XenoWars (changes requested):

packages/core/data/oot/world_mq/fire_temple_mq.yml
eedefeed: So replace:
`is_adult`

With:
`(is_adult || can_hookshot || climb_anywhere)`

Makes sense. I'm pretty sure child hookshot, using equip swap, is how I rescued a seed that bottlenecked at this check. I've not played with climb anywhere, so I can't tell you it definitely will work; if you'd require verification.

I do note that climb_anywhere isn't anywhere in the fire_mq file. Is there concern that checking for `climb_anywhere` might give a future editor a false impression that someone had already completely accounted for climb anywhere within this dungeon?

(Apologies for the tardiness in my reply, I've been away from home over the Christmas period)

XenoWars: considering im the one that did all the climb/hook anywhere and ageless logic, that shouldn't be an issue. i know i didnt do any mq for those.

eedefeed:
I think I've added changes that take into account @XenoWars's recommendations, but I don't _really_ understand git so I'm not sure. The 'Files changed' tab seems to be showing what I intended to show.

XenoWars:
Fix the conflicts with the file

eedefeed:
Is it obvious I have no idea what I'm doing?

(I think the conflict happened because of a big refactor Nax just pushed?)

PR #600, opened 2023-12-28
Shuffle the Zora Domain Diving Game Rupees. 

ZoeyZolotova:
Also add pond fish to the allsanity and hell presets and to mystery settings.

Not sure what to do for the 500 rupee orange rupee in the diving game, so for now I've made it be a Huge Rupee (200).

Also I get an error when running `npm install` so I'm letting github run the tests for me.


packages/core/lib/combo/logic/world-transform.ts
```diff
@@ -1035,6 +1035,22 @@ export class LogicPassWorldTransform {
       this.removeLocations(['OOT Hatch Chicken', 'OOT Hatch Pocket Cucco']);
     }
 
+    /* Handle Diving Game Rupees */
+    if (!settings.divingGameRupeeShuffle) {
```
Nax, line 1039: This should go before `makePools()` is called, that way you don't have to remove items, only locations (see how grass & pots handle that)

packages/core/src/oot/actors/En/En_Ex_Ruppy.S
```diff
@@ -0,0 +1,37 @@
...
+  lh      t0, 0x294(v0)
+  addiu   t1, t0, 0x1
+  jal     Actor_Kill
+  sh      t1, 0x294(v0)
+*/
+PATCH_START 0x80a8b5c8
+  or      a0, s0, zero
+  lw      a1, 0x0054 (sp)
+  jal     EnExRuppy_GiveItem
+  or      a2, v0, zero
```
Nax, line 29: Please follow the common MIPS convention of off-indenting by one space instructions in a branch delay slot

packages/core/data/oot/world/overworld.yml
```diff
@@ -1429,6 +1429,11 @@
   locations:
     "Zora Domain Waterfall Chest": "is_child"
     "Zora Domain Diving Game": "is_child && can_use_wallet(1) && soul_npc(SOUL_NPC_ZORA)"
+    "Zora Domain Diving Game Green Rupee": "is_child && can_use_wallet(1) && soul_npc(SOUL_NPC_ZORA)"
+    "Zora Domain Diving Game Blue Rupee": "is_child && can_use_wallet(1) && soul_npc(SOUL_NPC_ZORA)"
+    "Zora Domain Diving Game Red Rupee": "is_child && can_use_wallet(1) && soul_npc(SOUL_NPC_ZORA)"
+    "Zora Domain Diving Game Purple Rupee": "is_child && can_use_wallet(1) && soul_npc(SOUL_NPC_ZORA)"
+    "Zora Domain Diving Game Huge Rupee": "is_child && can_use_wallet(1) && soul_npc(SOUL_NPC_ZORA)"
```
Nax, line 1436: This would be better extracted to an area dedicated to the diving game (but not urgent)

packages/core/src/oot/actors/En/En_Ex_Ruppy.S
ZoeyZolotova: Like this
```c
  jal     EnExRuppy_GiveItem
   or      a2, v0, zero
  ```

or like this?
```c
  jal     EnExRuppy_GiveItem
   or     a2, v0, zero
```

Nax:
Otherwise looks pretty good

packages/core/lib/combo/logic/world-transform.ts
ZoeyZolotova: Done :)

packages/core/src/oot/actors/En/En_Ex_Ruppy.S
ZoeyZolotova: Done

packages/core/data/oot/world/overworld.yml
ZoeyZolotova: Noted for next time. Would the same ideally be done for the fishing pond fish and rewards?

Nax: Every time it avoids excessive repeats essentially. But yeah, not urgent, we still have a lot of these repeats.

Nax:
Looks good, will merge once tests pass

PR #625, opened 2024-01-23
Add ability to use Din's Fire, Farore's Wind and Nayru's Love in MM.

ZoeyZolotova:


packages/data/src/world/mm/overworld.yml
```diff
@@ -1392,11 +1392,10 @@
   exits:
     "Swamp Back": "true"
     "Deku Palace Cliff": "has(MASK_DEKU) || short_hook_anywhere"
-    "Near Deku Shrine": "event(CLEAN_SWAMP)"
+    "Near Deku Shrine": "event(CLEAN_SWAMP) || can_use_nayru"
```
XenoWars, line 1395: should check for zora with nayru.

packages/data/src/world/mm/overworld.yml
```diff
@@ -1564,7 +1563,7 @@
 "Near Deku Shrine":
   region: DEKU_PALACE
   exits:
-    "Deku Palace Front": "event(CLEAN_SWAMP) || has(MASK_DEKU)"
+    "Deku Palace Front": "event(CLEAN_SWAMP) || has(MASK_DEKU) || can_use_nayru"
```
XenoWars, line 1566: i wouldn't have this here unless when you die after being spit out by the big octo, you can use nayru before getting spit out again

XenoWars (changes requested):

packages/data/src/world/mm/overworld.yml
ZoeyZolotova: Do you mean because zora is required for killing the big octo?

packages/data/src/world/mm/overworld.yml
ZoeyZolotova: You can use it before the big octo gets you, and you can also get in a lilipad and use it.

PR #642, opened 2024-02-04
Simplify Evan check (and possibly improve it) plus add short hookshot zora hall trick

eedefeed:
Evan's HP check requires you to enter Evan's room _before_ Great Bay has been put in clear state. (Evan is not in the room after GB is cleared.)

With various settings, such as pre-completed dungeons, it has become possible to enter Evan's room after GB has been cleared, but without the items you need to actually complete GBT: namely: zora mask. These settings make it possible to have the ability to enter Evan's room, but only after Evan isn't there, making it impossible to get the check.

The current logic attempts to deal with this problem by checking for various combinations of settings, and if these are detected, making sure that you have Zora Mask (or Farore's Wind). I think that checking for these various settings is redundant. The same requirements of making-sure-you-can-actually-enter-the-room-before-clear are just as applicable regardless of settings - except interior ER. In other words, there's no combination of settings (except interior ER) that allows you to get Evan's check without Zora Mask (or Farore's Wind). 

With interior ER, you can roll this room elsewhere and thus be able to get the check without Zora mask... or you could roll the room in its vanilla location and still need to meet the enter-before-clear requirements. It doesn't seem to be possible to design logic to account for this, so it's best to remain logically restrictive when interior ER is switched on. I somewhat suspect the existing logic may allow for unbeatable seeds when this room is rolled vanilla (or any other room in Zora Hall).

To make the intended logic clearer, I have used a macro. Currently this macro only checks for Zora's Mask, but there are tricks/glitches that can be added later that allow you to open this door without Zora Mask. If we didn't use a macro, then whoever adds the trick may add the trick to the entrance, but forget to add the trick to the Evan check itself.


packages/data/src/macros/macros_mm.yml
```diff
@@ -144,6 +144,7 @@
 "can_pass_gibdo": "has(MASK_GIBDO) && soul_redead_gibdo"
 "can_get_gossip_fairy": "can_play_healing || can_play_epona"
 "can_jumpslash": "has_weapon || has_sticks || has_mask_zora"
+"can_enter_evans_before_gb_clear": "has_mask_zora"
```
XenoWars, line 147: Could you add the `short_hook_anywhere && trick(MM_ZORA_HALL_DOORS)` to this? Would need to go to core/lib/combo/settings/tricks.ts to add it. Description could be "Access the Doors in Zora Hall using Short Hookshot Anywhere" and then would also need added in core/lib/combo/preset.ts to the hell preset since it uses every trick.

I did confirm it works on all four doors

XenoWars (changes requested):

XenoWars:
PR on hold until a decision is made on how the logic should be handled.

PR #645, opened 2024-02-05
MM: Iron+Hover Boots Logic Refining

XenoWars:
This logic goes through and refines the Iron and Hover Boots logic in various regions within MM, including the four main dungeons of the game. Also adds various tricks in accordance with some of the logic.

And most importantly, more Stone Mask and Bunny Hood logic!


packages/data/src/world/mm/great_bay_temple.yml
```diff
@@ -44,7 +44,7 @@
     MAGIC: "true"
   locations:
     "Great Bay Temple SF Central Room Barrel": "true"
-    "Great Bay Temple SF Central Room Underwater Pot": "has_mask_zora || (has_arrows && (has(MASK_GREAT_FAIRY) || underwater_walking_strict))"
+    "Great Bay Temple SF Central Room Underwater Pot": "has_mask_zora || (has(MASK_GREAT_FAIRY) && (has_arrows || (underwater_walking_strict && has_mask_blast)))"
```
ZoeyZolotova, line 47: This isn't right. You dont need great fairy's mask to get the fairy with iron boots, as long as you can break the pot somehow. So overall it should be GFM + Arrows OR Underwater Walking + Blast Mask OR Underwater Walking + Arrows

packages/data/src/world/mm/great_bay_temple.yml
```diff
@@ -105,7 +104,7 @@
     "Great Bay Temple Central Room": "underwater_walking || hookshot_anywhere"
     "Great Bay Temple Ice Arrow Room": "small_keys_gb(1)"
   events:
-    GB_PIPE_RED: "(soul_octorok && can_use_ice_arrows) || short_hook_anywhere"
+    GB_PIPE_RED: "(soul_octorok && can_use_ice_arrows) || short_hook_anywhere || (has_hover_boots && has_mask_bunny && has_weapon && trick(MM_GBT_RED1_HOVERS))"
```
ZoeyZolotova, line 107: How is this done? From the barrel/crate or something?

ZoeyZolotova, line 107: Is this running from the red pipe to the platform with the switch? Does it need a jump slash and bunny or can it be done with a backwalk + backflip?

packages/data/src/world/mm/snowhead_temple.yml
```diff
@@ -206,10 +206,10 @@
   exits:
     "Snowhead Temple Map Room Upper": "true"
     "Snowhead Temple Center Level 2 Dual": "true"
-    "Snowhead Temple Center Level 3 Iced": "has_mask_goron || can_hookshot || short_hook_anywhere || has_hover_boots" # todo hover boots trick
+    "Snowhead Temple Center Level 3 Iced": "has_mask_goron || can_hookshot || short_hook_anywhere || has_hover_boots" # Hovers trick not necessary until other settings call for it
     "Snowhead Temple Snow Room": "small_keys_sh(3)"
     "Snowhead Temple Fire Arrow": "true"
-    "Snowhead Temple Center Level 4": "short_hook_anywhere && (can_use_fire_short_range || has_mask_blast || has_bombchu)"
+    "Snowhead Temple Center Level 4": "short_hook_anywhere && (can_use_fire_arrows || has_mask_blast || has_bombchu)"
```
ZoeyZolotova, line 212: Is this shooting the snowballs with fire arrows and then hookshotting to the area? does din's not reach?

packages/data/src/world/mm/great_bay_temple.yml
XenoWars: You can't get the fairy even with iron boots. i tried and tried and could not get it despite all the rolling towards the center i did.

If it's possible to get, should most likely be behind a trick.

packages/data/src/world/mm/great_bay_temple.yml
XenoWars: This is running from Wart's room. the pipe is too low.

packages/data/src/world/mm/snowhead_temple.yml
XenoWars: din's doesn't reach no. i tried at the very corner.

packages/data/src/world/mm/great_bay_temple.yml
ZoeyZolotova: I did try this and it is possible. Definitely annoying though.

PR #652, opened 2024-02-06
Add Silver Scale and Golden Scale to MM.

ZoeyZolotova:


packages/data/src/macros/macros_common.yml
```diff
@@ -9,6 +9,8 @@
 "has_tunic_goron_raw": "cond(setting(sharedTunicGoron), has(SHARED_TUNIC_GORON), has(TUNIC_GORON))"
 "has_tunic_zora_raw": "cond(setting(sharedTunicZora), has(SHARED_TUNIC_ZORA), has(TUNIC_ZORA))"
 
+"has_scale_raw(x)": "cond(setting(sharedTunicZora), has(SHARED_SCALE, x), has(SCALE, x))"
```
XenoWars, line 12: I would imagine this should be Shared Scale, not Tunic.

packages/data/src/macros/macros_mm.yml
```diff
@@ -31,6 +31,10 @@
 "has_tunic_goron": "has_tunic_goron_strict || trick(MM_TUNICS)"
 "has_tunic_zora": "has_tunic_zora_strict || trick(MM_TUNICS)"
 
+# Scales
+"can_dive_small": "has_scale_raw(1)"
+"can_dive_big": "has_scale_raw(2)"
```
XenoWars, line 36: Could maybe include ` || underwater_walking` with these, like how OOT's include Iron Boots. Though would then need to remove underwater_walking in every or clause that has them both.

packages/data/src/world/mm/great_bay_temple.yml
```diff
@@ -52,7 +52,7 @@
   events:
     MAGIC: "true"
   exits:
-    "Great Bay Temple Baba Room": "has_mask_zora || ((has_bombchu || has_mask_blast) && (short_hook_anywhere || underwater_walking)) || hookshot_anywhere"
+    "Great Bay Temple Baba Room": "has_mask_zora || ((has_bombchu || has_mask_blast) && (short_hook_anywhere || underwater_walking || can_dive_big)) || hookshot_anywhere"
```
XenoWars, line 55: This will require zora tunic as it takes too much time to reach the room left as is. If it is extended to a 300 frame delay on the timer, then it would be fine, but i'd still say it should at least soft require the tunic.

packages/data/src/world/mm/great_bay_temple.yml
```diff
@@ -94,9 +94,9 @@
     "Great Bay Temple Pot Compass Room Surface 2": "true"
     "Great Bay Temple Pot Compass Room Surface 3": "true"
     "Great Bay Temple Pot Compass Room Surface 4": "true"
-    "Great Bay Temple Pot Compass Room Water 1": "underwater_walking || short_hook_anywhere"
-    "Great Bay Temple Pot Compass Room Water 2": "underwater_walking || short_hook_anywhere"
-    "Great Bay Temple Pot Compass Room Water 3": "underwater_walking || short_hook_anywhere"
+    "Great Bay Temple Pot Compass Room Water 1": "underwater_walking || short_hook_anywhere || (can_dive_small && (has_weapon_range || has_bombchu))"
+    "Great Bay Temple Pot Compass Room Water 2": "underwater_walking || short_hook_anywhere || (can_dive_small && (has_weapon_range || has_bombchu))"
+    "Great Bay Temple Pot Compass Room Water 3": "underwater_walking || short_hook_anywhere || (can_dive_small && (has_weapon_range || has_bombchu))"
```
XenoWars, line 99: these should check specifically for arrows or hook since deku bubbles just splat on the water's surface.

packages/data/src/world/mm/great_bay_temple.yml
```diff
@@ -180,14 +180,14 @@
...
+    "Great Bay Temple Pot Green Pipe 2 3": "(can_dive_small && (has_weapon_range || has_bombchu)) || underwater_walking || hookshot_anywhere || (can_use_ice_arrows && short_hook_anywhere)"
+    "Great Bay Temple Pot Green Pipe 2 4": "(can_dive_small && (has_weapon_range || has_bombchu)) || underwater_walking || hookshot_anywhere || (can_use_ice_arrows && short_hook_anywhere)"
+    "Great Bay Temple Pot Green Pipe 2 5": "(can_dive_small && (has_weapon_range || has_bombchu)) || underwater_walking || hookshot_anywhere || (can_use_ice_arrows && short_hook_anywhere)"
+    "Great Bay Temple Pot Green Pipe 2 6": "(can_dive_small && (has_weapon_range || has_bombchu)) || underwater_walking || hookshot_anywhere || (can_use_ice_arrows && short_hook_anywhere)"
+    "Great Bay Temple Pot Green Pipe 2 7": "(can_dive_small && (has_weapon_range || has_bombchu)) || underwater_walking || (can_use_ice_arrows && short_hook_anywhere)"
+    "Great Bay Temple Pot Green Pipe 2 8": "(can_dive_small && (has_weapon_range || has_bombchu)) || underwater_walking || (can_use_ice_arrows && short_hook_anywhere)"
```
XenoWars, line 190: All of these will need to check specifically for arrows or hook, the last two only arrows since hook cannot reach them.

packages/data/src/world/mm/great_bay_temple.yml
```diff
@@ -230,10 +230,10 @@
...
     "Great Bay Temple Boss Pot 3": "true"
     "Great Bay Temple Boss Pot 4": "true"
-    "Great Bay Temple Boss Pot Underwater 1": "underwater_walking"
-    "Great Bay Temple Boss Pot Underwater 2": "underwater_walking"
-    "Great Bay Temple Boss Pot Underwater 3": "underwater_walking"
-    "Great Bay Temple Boss Pot Underwater 4": "underwater_walking"
+    "Great Bay Temple Boss Pot Underwater 1": "underwater_walking || (has_mask_blast && can_dive_big)" # todo blast+dive should be a trick
+    "Great Bay Temple Boss Pot Underwater 2": "underwater_walking || (has_mask_blast && can_dive_big)" # todo blast+dive should be a trick
+    "Great Bay Temple Boss Pot Underwater 3": "underwater_walking || (has_mask_blast && can_dive_big)" # todo blast+dive should be a trick
+    "Great Bay Temple Boss Pot Underwater 4": "underwater_walking || (has_mask_blast && can_dive_big)" # todo blast+dive should be a trick
```
XenoWars, line 236: Not to mention the logic could have you do these without having the means to defeat Gyrog... You are definitely not getting these while Gyorg is present. These with scale should be placed in the After Boss region so that you can defeat Gyorg with FD and then get them.

packages/data/src/world/mm/overworld.yml
```diff
@@ -2270,7 +2270,7 @@
 "Great Bay Coast Fortress":
   region: GREAT_BAY_COAST
   exits:
-    "Great Bay Coast": "underwater_walking"
+    "Great Bay Coast": "underwater_walking || can_dive_small"
```
XenoWars, line 2273: Not sure how you're gonna get past the wooden plank, especially since you can't even dive here anyway due to being entirely underwater.

packages/data/src/world/mm/overworld.yml
```diff
@@ -2517,12 +2517,12 @@
 "Zora Cape Near Hall":
   region: ZORA_CAPE
   exits:
-    "Zora Cape": "underwater_walking"
-    "Zora Hall Entrance": "underwater_walking"
+    "Zora Cape": "underwater_walking || can_dive_small"
```
XenoWars, line 2520: I feel like this should also require the human trick if the way here uses it. Of course, none of it is relevant until ER shuffles the entrance.

packages/data/src/world/mm/stone_tower_temple.yml
```diff
@@ -72,7 +72,7 @@
 "Stone Tower Temple Center":
   dungeon: ST
   exits:
-    "Stone Tower Temple Water Room": "underwater_walking || short_hook_anywhere"
+    "Stone Tower Temple Water Room": "underwater_walking || short_hook_anywhere || can_dive_small"
```
XenoWars, line 75: I feel like the Dexihand would be getting in your way, no? Should probably include methods to defeat it (does not need the soul)

packages/data/src/world/mm/stone_tower_temple.yml
```diff
@@ -83,7 +83,7 @@
 "Stone Tower Temple Water Room":
   dungeon: ST
   exits:
-    "Stone Tower Temple Center": "has_mask_zora || (soul_enemy(SOUL_ENEMY_DEXIHAND) && (short_hook_anywhere || (has_tunic_zora && has_iron_boots))) || (has_tunic_zora && has_iron_boots && (can_hookshot || short_hook_anywhere))"
+    "Stone Tower Temple Center": "has_mask_zora || (soul_enemy(SOUL_ENEMY_DEXIHAND) && (short_hook_anywhere || can_dive_small || (has_tunic_zora && has_iron_boots))) || (has_tunic_zora && has_iron_boots && (can_hookshot || short_hook_anywhere))"
```
XenoWars, line 86: gonna need to know how much timer expires when doing this and the above one. If it's more than 34s with the 80 frames, it's gonna need to hard require tunic, otherwise should probably include tunic if it's too close to 34s like with GBT's Baba room.

XenoWars (changes requested):
Logic review

packages/core/src/common/text/text.c
```diff
@@ -711,7 +717,9 @@ void comboTextAppendItemNameEx(char** b, s16 gi, int flags, int importance)
             itemName = "a " TEXT_C2 "Progressive Goron Lullaby";
             break;
         case GI_OOT_SCALE_SILVER:
+        case GI_MM_SCALE_SILVER:
             itemName = "a " TEXT_C1 "Progressive Scale";
+            ambiguous = !comboConfig(CFG_SHARED_SCALES);
```
Nax, line 722: Needs to check for the active setting too (it's not ambiguous if MM scale isn't on)

packages/data/src/macros/macros_mm.yml
ZoeyZolotova: I was considering that, but decided not to do that. I like it how it is.

packages/data/src/world/mm/great_bay_temple.yml
ZoeyZolotova: No they dont.

packages/data/src/world/mm/great_bay_temple.yml
ZoeyZolotova: Ah yeah, or ice arrows.

packages/data/src/world/mm/great_bay_temple.yml
ZoeyZolotova: Sounds good, but if you come here without gyorg's soul, the pots would be out of logic using a dive. is that okay?

packages/data/src/world/mm/overworld.yml
ZoeyZolotova: I thought this was referring to leaving pirates fortress. Reverted.

packages/data/src/world/mm/stone_tower_temple.yml
ZoeyZolotova: The dexihand is easily avoided with the silver scale.

packages/data/src/world/mm/stone_tower_temple.yml
ZoeyZolotova: It takes less than 10 seconds from when the timer begins to get through the tunnel.

packages/data/src/world/mm/great_bay_temple.yml
```diff
@@ -179,14 +179,14 @@
...
+    "Great Bay Temple Pot Green Pipe 2 4": "(can_dive_small && (has_arrows || has_bombchu || can_hookshot)) || underwater_walking || hookshot_anywhere || (can_use_ice_arrows && (short_hook_anywhere || can_dive_small))"
+    "Great Bay Temple Pot Green Pipe 2 5": "(can_dive_small && (has_arrows || has_bombchu || can_hookshot)) || underwater_walking || hookshot_anywhere || (can_use_ice_arrows && (short_hook_anywhere || can_dive_small))"
+    "Great Bay Temple Pot Green Pipe 2 6": "(can_dive_small && (has_arrows || has_bombchu || can_hookshot)) || underwater_walking || hookshot_anywhere || (can_use_ice_arrows && (short_hook_anywhere || can_dive_small))"
+    "Great Bay Temple Pot Green Pipe 2 7": "(can_dive_small && (has_arrows || has_bombchu)) || underwater_walking || (can_use_ice_arrows && (short_hook_anywhere || can_dive_small))"
+    "Great Bay Temple Pot Green Pipe 2 8": "(can_dive_small && (has_arrows || has_bombchu)) || underwater_walking || (can_use_ice_arrows && (short_hook_anywhere || can_dive_small))"
```
XenoWars, line 189: Why exactly does it need to be with ice arrows if arrows alone are already enough?

XenoWars (changes requested):

PR #655, opened 2024-02-07
Show icons of next item on a slot if there is any.

ZoeyZolotova:
This also fixes item slots becoming empty when attempting to cycle them when they contain an item that doesn't belong.


packages/core/src/mm/ovl/kaleido_scope.c
```diff
@@ -436,22 +400,102 @@ void KaleidoScope_LoadIcons(u32 vrom, void* dst, size_t* size)
...
+
+    return dstVtx;
+}
+
+static void DrawIcon(GfxContext* gfxCtx, const Vtx* vtx, u32 segAddr, u16 width, u16 height, u16 qidx) {
+    OPEN_DISPS(gfxCtx);
+    // Instructions that happen before function
+    gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, 0xFF, 0xFF, 0xFF, gfxCtx->play->pauseCtx.itemAlpha & 0xFF);
+    gSPVertex(POLY_OPA_DISP++, vtx, 4, 0); // Loads 4 vertices from RDRAM
+    // Instructions that happen during function.
```
Nax, line 472: This should use `gDPLoadTexture`

packages/core/src/oot/ovl/kaleido_scope.c
```diff
@@ -1552,46 +1538,124 @@ s32 KaleidoScope_BeforeDraw(GameState_Play* play)
...
+
+    return dstVtx;
+}
+
+static void DrawIcon(GfxContext* gfxCtx, const Vtx* vtx, u32 segAddr, u16 width, u16 height, u16 qidx) {
+    OPEN_DISPS(gfxCtx);
+    // Instructions that happen before function
+    gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, 0xFF, 0xFF, 0xFF, gfxCtx->play->pauseCtx.alpha & 0xFF);
+    gSPVertex(POLY_OPA_DISP++, vtx, 4, 0); // Loads 4 vertices from RDRAM
+    // Instructions that happen during function.
```
Nax, line 1624: Ditto

PR #657, opened 2024-02-08
Add settings that change the number of skulltula tokens and stray fai…

Platypus-Gamer:
…ries needed to get their rewards
This adds 3 settings- one to change the number of stray fairies needed to get the great fairy rewards, one to do the same for Skulltula tokens and the rewards at the houses, and one to choose whether they should be included in logic or not. The first two settings allow players to pick from multiple options that each correspond to a fairy/skull number respectively. I'm not too sure if the last setting is necessary, my intention was for this to help more casual players who have difficulty finding all of the fairies to have a buffer so they can miss a few without logic punishing them but I don't know if that could lead to some bad accidental logic breaking. It also adds the implementation for these features that work properly for me and should work properly everywhere.
This is my first time trying to commit here, hopefully, these changes work and can be added to the rando! Please give me any feedback on things I missed or need to change. Thank you!
https://github.com/OoTMM/OoTMM/issues/511


CelestialKitsune:
Ultimately, as mentioned on the Discord, we'd change this to special conditions for all of them. But for that we need to refactor the web component first

Nax:
We might not want the full flexibility of special conditions for these. I think a numeric value that can be randomized (like prices) would be a good compromise, when we get everything to support it.

Can't review right now but I'll do when I have a bit more time.

Platypus-Gamer:
> We might not want the full flexibility of special conditions for these. I think a numeric value that can be randomized (like prices) would be a good compromise, when we get everything to support it.

I agree with this. I changed the skull and fairy counts to be numbers instead of enums and integrated them into the code, still unsure about the logic. Will probably add random/weighted random options and incorporate this into random settings and spoiler log for now.

PR #660, opened 2024-02-08
Add Strength Upgrades for MM. 

ZoeyZolotova:
Need to tweak the gauntlets to fit MM Link better.


packages/data/src/world/mm/snowhead_temple.yml
```diff
@@ -167,26 +167,34 @@
...
     "Snowhead Temple Rupee 1": "has_arrows"
     "Snowhead Temple Rupee 2": "has_arrows"
     "Snowhead Temple Rupee 3": "has_arrows"
-"Snowhead Temple Dual Switches":
+"Snowhead Temple Dual Switches Locked":
   dungeon: SH
   exits:
     "Snowhead Temple Icicles": "small_keys_sh(3) || ((has_explosives || trick_keg_explosives) && small_keys_sh(2))"
-    "Snowhead Temple Center Level 2 Dual": "true"
+    "Snowhead Temple Dual Switches Unlocked": "has_mask_goron || has_mask_zora || short_hook_anywhere || has_hover_boots || can_use_fire_short_range || trick_sht_fireless_er"
```
XenoWars, line 181: Could be wrong, but aren't there crates on the other side you can hookshot to?

packages/data/src/world/mm/snowhead_temple.yml
```diff
@@ -241,8 +249,8 @@
     "Snowhead Temple Center Level 3 Snow": "true"
     "Snowhead Temple Center Level 4": "(trick(MM_SHT_PILLARLESS) && (can_use_fire_arrows || has_bombs)) || (short_hook_anywhere && (has_bombs || can_use_fire_arrows)) || hookshot_anywhere"
   locations:
-    "Snowhead Temple SF Dinolfos 1": "soul_lizalfos_dinalfos"
-    "Snowhead Temple SF Dinolfos 2": "soul_lizalfos_dinalfos"
+    "Snowhead Temple SF Dinolfos 1": "soul_lizalfos_dinalfos && (can_fight || has_arrows)"
+    "Snowhead Temple SF Dinolfos 2": "soul_lizalfos_dinalfos && (can_fight || has_arrows)"
```
XenoWars, line 253: We have Deku Bubbles in logic for them on the Moon too, although that one is only one.

XenoWars (changes requested):
Can't put it on the specific areas, but should check for methods to defeat the Wizzrobes in Snowhead Temple. They both assume you have Foron or Zora, but it will be possible to get to them without a means to fight.

packages/data/src/world/mm/pirate_fortress.yml
```diff
@@ -26,19 +26,34 @@
...
-    "Pirate Fortress Sewers Rupee Near Cage 2": "has_mask_zora"
-    "Pirate Fortress Sewers Rupee Near Cage 3": "has_mask_zora"
-    "Pirate Fortress Sewers Rupee Near Cage 4": "has_mask_zora"
-    "Pirate Fortress Sewers Rupee Cage": "has_mask_zora"
+"Pirate Fortress Sewers Cage Room":
+  region: PIRATE_FORTRESS_SEWERS
+  exits:
+    "Pirate Fortress Entrance": "true"
+    "Pirate Fortress Sewers": "has_mask_zora || (underwater_walking_strict && can_lift_silver)"
+    "Pirate Fortress Sewers After Gate": "has_weapon_range || (has_hover_boots && (has_sticks || has_weapon))"
```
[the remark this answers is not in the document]
ZoeyZolotova: Hovers do make it. I don't want to assume goron for it though, because the sewer could be added the entrance rando.

packages/data/src/world/mm/snowhead_temple.yml
ZoeyZolotova: You're right. Does short hookshot reach the crates?

packages/data/src/world/mm/snowhead_temple.yml
ZoeyZolotova: I used the same logic as for the dinolfos in woodfall temple. Should they all change to account for deku bubble?

packages/data/src/world/mm/snowhead_temple.yml
ZoeyZolotova: It doesn't, but yeah long hookshot should be in logic for it.

packages/data/src/world/mm/pirate_fortress.yml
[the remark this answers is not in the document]
XenoWars: So Goron should probably be included in the possibilities then

packages/data/src/world/mm/snowhead_temple.yml
XenoWars: WFT one could. but since there's two in this room, it'd probably be too risky of getting burned.

XenoWars (approved):
Logic is good

PR #662, opened 2024-02-10
Updated stray fairy and MM skulltula requirement settings

Platypus-Gamer:
I took a stab at updating the Great Fairy and Spider House reward settings and I think this works. It can now be any number or a random number (or a weighted random but idk how to do that well) and might have functional logic, but I'm not sure. I tried copying shop-sanity's logic system which doesn't crash things but might not actually behave as functional logic which I am not sure how to test. The number selector and random number options are both functional, so I will do some testing on the logic to see if it works. Again, let me know any changes or fixes that I need to make, thank you!


Platypus-Gamer:
Ok the logic definitively does not work, it treats it as if you always have enough stray fairies/skull tokens. I've tried a lot of things to fix this but nothing has worked (in fact they may have just made things worse), so I would appreciate it if other people could give it a glance 👍

XenoWars:
Closing this due to mass amount of conflicts and there's been no word for almost three months.

PR #676, opened 2024-02-18
Lost Woods Exits

ZoeyZolotova:
Add a setting to alter the "getting lost" lost woods exits to take you into the lost woods again. Add a setting to randomize those entrances.

### DIFF
--- a/packages/core/lib/combo/settings/data.ts
+++ b/packages/core/lib/combo/settings/data.ts
@@ -1955,7 +1955,7 @@ export const SETTINGS = [{
   name: 'Shuffle Child Owl Flights',
   category: 'entrances',
   type: 'boolean',
-  description: 'Shuffle the two owl flights among one-way entrances.',
+  description: 'Shuffle the Lost Woods entrances added by "Alter Lost Woods Exits" among one-way entrances.',
   default: false,
   cond: (x: any) => x.erOneWays !== 'none' && x.alterLostWoodsExits
 }] as const;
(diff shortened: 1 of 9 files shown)


packages/core/lib/combo/settings/data.ts
```diff
@@ -1943,4 +1950,12 @@ export const SETTINGS = [{
   description: 'Shuffle the two owl flights among one-way entrances.',
   default: false,
   cond: (x: any) => x.erOneWays !== 'none'
+}, {
+  key: 'erOneWaysWoods',
+  name: 'Shuffle Child Owl Flights',
```
Nax, line 1955: Copy-paste error

Nax (changes requested):
Need to name the ER correctly, otherwise good

packages/data/src/world/oot/overworld.yml
```diff
@@ -799,6 +815,10 @@
     "Goron City Shortcut": "true"
     "Zora River": "can_dive_small || hookshot_anywhere"
     "Lost Woods Rupee Arrow": "is_child && (can_dive_small || hookshot_anywhere || can_boomerang)"
+    "Lost Woods Lost North": "true"
+    "Lost Woods Lost East": "true"
+    "Lost Woods Lost South": "true"
+    "Lost Woods Lost West": "true"
```
XenoWars, line 821: You cannot access East in the first part of Lost Woods (before Mido). East can only be accessed in Lost Woods Deep. All four should be added to Lost Woods Deep while doing since adult can't access the ones in the front if coming from SFM

XenoWars (changes requested):

XenoWars:
Should also be added to the "anyEr" in "private runAttempt" in entrance.ts

PR #678, opened 2024-02-19
Add note about logic changes on New Cycle.

bndbsh:
It's not clear otherwise that this has logic implications.


XenoWars:
This isn't needed. That's what we have a FAQ for that lists all the methods that MM is in logic.

bndbsh:
What FAQ are you talking about? From [the website](https://ootmm.com/faq):

> **Can the MM Ocarina / MM Song of Time be found in Majora's Mask?**
> No, the logic expects you to have the Ocarina and Song of Time before doing any check in Majora's Mask. In other words, the MM Ocarina and MM Song of Time will always be placed in Ocarina of Time.

PR #690, opened 2024-03-12
Add Adult Link in MM.

ZoeyZolotova:


packages/core/include/combo/common/api.h
```diff
@@ -119,6 +119,8 @@ void DmaCompressed(u32 pstart, void* dst, u32 size);
 #if defined(GAME_OOT)
 void _DmaCompressed(u32 pstart, void* dst, u32 size);
 #endif
+s32 RequestDma(DmaRequest* request, void* vramStart, u32 vromStart, size_t size, u32 unused,
+                           OSMesgQueue* queue, OSMesg msg);
```
Nax, line 123: Style: single line

packages/core/include/combo/mm/object.h
```diff
@@ -0,0 +1,31 @@
+#ifndef INCLUDED_COMBO_MM_OBJECT_H
+#define INCLUDED_COMBO_MM_OBJECT_H
+
+#include <assert.h>
+#include <combo/util.h>
+
+#define OBJECT_EXCHANGE_BANK_MAX 35
+
+typedef struct {
```
Nax, line 9: * Braces on their own lines
* No offset comments - use assertions instead
* No size comment, use a size assertion

packages/core/lib/combo/settings/data.ts
```diff
@@ -881,6 +881,13 @@ export const SETTINGS = [{
   type: 'boolean',
   description: 'Makes the Mask Shop in Market open during the night',
   default: false
+}, {
+  key: 'crossAge',
+  name: 'Cross-Games Age',
+  category: 'main.cross',
+  type: 'boolean',
+  description: 'When you enter MM as Adult Link, you will be Adult Link in MM. Model credit: Skilar.',
```
Nax, line 889: Credits shouldn't be here, or else people are gonna start adding credits for every asset/texture/logic/whatnot and it'll get unreadable (see old BSD software where the credits is larger than the documentation).

The model file should go in its own commit, commited by you but authored by Skilar, or mentionned in the commit descr, or both.

Once we add in-game credits we can list him there too.

packages/core/src/mm/scene.c
```diff
@@ -20,3 +21,145 @@ PATCH_CALL(0x8013231c, Entrance_GetTableEntryCustom);
...
+            }
+        }
+
+        entry++;
+    }
+}
+
+PATCH_FUNC(0x8012F4FC, Object_UpdateEntriesCustom);
+
+void Object_AfterInitContext()
```
Nax, line 162: Empty params list and `(void)` are not the same in C, this needs to be `(void)`

packages/core/src/mm/scene.c
```diff
@@ -20,3 +21,145 @@ PATCH_CALL(0x8013231c, Entrance_GetTableEntryCustom);
...
+
+    addr = ((u32)objectCtx->slots[slot].segment) + vromSize;
+    addr = ALIGN16(addr);
+
+    return (void*)addr;
+}
+
+PATCH_FUNC(0x8012f73c, Object_AllocateSlotCustom);
+
+void Object_UpdateEntriesCustom(ObjectContext* objectCtx)
```
Nax, line 122: Not sure I understand the point of this function, isn't it pretty much what the vanilla game is doing?

Nax (changes requested):

packages/core/src/mm/scene.c
ZoeyZolotova: It's pretty much the same except that instead of `objectFile = &gObjectTable[id];` we use `objectFile = comboGetObjectData(id);` so that non-vanilla objects can work with this function.

packages/core/lib/combo/settings/data.ts
ZoeyZolotova: Ok I will amend the commits.

PR #750, opened 2024-07-15
Add setting for Grass/Pot CAMC

XenoWars:
Separates the Grass and Pot CAMC out from the rest and makes it its own setting, still affected by Stone of Agony if enabled. This enables the ability to have the usual CAMC on without altering the behavior of all grass and pots being gold until checked, or having only Grass and Pot CAMC.


dilijev:
Is CAMC a typo of CSMC (which seems more common?)

XenoWars:
> Is CAMC a typo of CSMC (which seems more common?)
> 
> (Apologies if this is a dumb question - I'm new here but I'm interested in contributing!)

Not a typo. Container Appearance Matches Contents. We dont separately alter the size and texture, so it makes more sense to call it more accurate to what it does.

CelestialKitsune:
I would recommend splitting it per container instead of coupling them. It would be inconsistent to have groups of specific containers.

PR #754, opened 2024-07-17
Fire temple access trick

ScienceNerd2240:
same as the other one, just refreshing the branch im using


ScienceNerd2240:
It is currently base logic to be able to access the Fire Temple as child when you have Hover Boots. This is kinda trivial in my opinion, and I thought it should be made into a trick. So I made it a trick. Thus the PR. I don't mind if we decide to keep it base logic, but I figured I'd at least make the trick and do the logic work for it.

CelestialKitsune:
I do not really see the point of adding this as a trick. You even say it is trivial, which means it's easy to pull off, and not really obscure.

ScienceNerd2240:
ok ill close the PR

PR #757, opened 2024-07-18
Deku Tree MQ Rework

Joshua8600:


packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,282 @@
...
-    "MQ Deku Tree Grass Room Before Spike 4": "can_cut_grass && (can_use_bow || (can_use_slingshot && (can_use_sticks || has_fire)))"
-    "MQ Deku Tree Heart Lobby": "true"
-    "MQ Deku Tree Heart Before Compass": "can_use_bow || can_use_sticks || can_use_din"
-    "MQ Deku Tree Heart Slingshot Room": "true"
-"Deku Tree Compass Room":
+"MQ Deku Tree Room Before Compass Room Entry Door":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree 2nd Floor": "true"
+    "MQ Deku Tree Room Before Compass Room Door to 2nd Floor": "true"
```
XenoWars, line 37: Do we know yet if this is actually the case? Ghun has been testing, so he should have been able to come through this door right? Are you behind the web?

packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,282 @@
...
     "MQ Deku Tree Grass Entrance Lower 5": "can_cut_grass"
+"MQ Deku Tree 2nd Floor":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Lobby": "true"
+    "MQ Deku Tree 3rd Floor": "true"
+    "MQ Deku Tree Room Before Compass Room Entry Door": "has_fire || (can_use_sticks && event(DEKU_MQ_MAIN_TORCH)) || (event(DEKU_MQ_MAIN_TORCH) && can_use_bow)"
+  locations:
+    "MQ Deku Tree Map Chest": "true"
+    "MQ Deku Tree GS Lobby Crate": "gs && can_damage_skull"
```
XenoWars, line 29: for now this is fine, but hook, bow, and slingshot would need to be able to break the crate.

packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,282 @@
...
     "MQ Deku Tree GS Compass Room": "gs && can_collect_distance"
     "MQ Deku Tree Heart Compass Room": "true"
-"Deku Tree Water Room":
+"MQ Deku Tree 3rd Floor":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree 2nd Floor": "true"
+    "MQ Deku Tree Lobby": "true"
+    "MQ Deku Tree Basement": "true"
+    "MQ Deku Tree Slingshot Room Entry Door": "has_fire || event(DEKU_MQ_MAIN_TORCH)"
```
XenoWars, line 100: Considering it's just a floor switch you walk over to and press, this can just be true.

packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,282 @@
...
+  exits:
+    "MQ Deku Tree Room Before Water Room Entry Door": "event(DEKU_MQ_ROOM_BEFORE_WATER_CLEAR)"
+    "MQ Deku Tree Room Before Water Room": "true"
+"MQ Deku Tree Room Before Water Room":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Room Before Water Room Door to Basement": "true"
+    "MQ Deku Tree Room Before Water Room Door to Water Room": "true"
+  events:
+    DEKU_MQ_WATER_PATH_TORCH1: "has_fire"
```
XenoWars, line 159: "cond(setting.doorShuffle !== none, has_fire, (roomClearEvent && has_sticks) || has_fire)"

This is meant to be here, not above. There isn't a reason to have the other event. And should include sticks back into the current logic since door rando is not yet a thing. Doesn't matter which direction you come from either as both directions have a torch you can carry you through.

packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,282 @@
...
+"MQ Deku Tree Water Room Door to Room Before Water Room":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Room Before Water Room Door to Water Room": "true"
+    "MQ Deku Tree Water Room Front": "true"
+"MQ Deku Tree Water Room Front":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Water Room Door to Room Before Water Room": "true"
+    "MQ Deku Tree Water Room Back": "true" #You can do this as both adult and child, without a shield, by rolling under the spike.
```
XenoWars, line 180: I'm fairly certain the strict requirements were a mercy. Yes, you can do it with a roll, but should it be expected when it can be finnicky? Probably not. And OOTR also has it setup that way with the roll as a trick. This also goes for the event below.

packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,282 @@
...
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Room After Water Room Door to Gravestone Room": "true"
+    "MQ Deku Tree MQ Grave Room": "true"
+"MQ Deku Tree MQ Grave Room":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Gravestone Room Door to Room After Water Room": "true"
+    "MQ Deku Tree Back Side Room Entry Door": "can_use_sticks || has_fire"
+    "MQ Deku Tree Basement Ledge": "is_child && (can_use_sticks || can_use_din)"
```
XenoWars, line 234: i'd make an event for this web blocking the crawlspace. Very possible you could use bow to burn it then come back as child.. Likewise, bow alone is fine for the side room (but does not need an event). So both can just be `can_use_sticks || can_use_din || can_use_bow`
Unless child is too short to just move to the side and shoot, then he would need elegy or fire arrows

packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,282 @@
...
+    "MQ Deku Tree Back Side Room Exit Door": "true"
+"MQ Deku Tree Back Side Room Exit Door":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Back Side Room Entry Door": "true"
+    "MQ Deku Tree Back Side Room": "true"
+"MQ Deku Tree Back Side Room":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Back Side Room Exit Door": "true" #One-sided collision for webbing; might need to be changed for door rando.
```
XenoWars, line 262: If the web is visible, it needs to use the trick. Otherwise, this is fine. (Same for the other webs in the main room if you spawn behind them)

packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,282 @@
...
+    "MQ Deku Tree Grass Back Room 1": "can_cut_grass"
+    "MQ Deku Tree Grass Back Room 2": "can_cut_grass"
+    "MQ Deku Tree Grass Back Room 3": "can_cut_grass"
+"MQ Deku Tree Basement Ledge":
   dungeon: DT
   exits:
-    "Deku Tree Before Boss": "can_use_sticks || has_fire"
-    "Deku Tree Backrooms": "is_child"
+    "MQ Deku Tree Basement": "true"
+    "Deku Tree Before Boss": "(can_use_sticks && (is_adult || (is_child && event(DEKU_BLLOCK)))) || has_fire"
```
XenoWars, line 275: child would be able to use hover boots for this too. Also nice typo

XenoWars (changes requested):

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: Not true. In MQ, spiders are under crates, not spawned from breaking them. If you kill this spider, item appears on top of crate.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: it is "true" via the event itself being true. I don't want to have the whole thing as true if eventual actorizer or something affecting it.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: The roll is not finnicky. I stand by this; if Outside Ganon Castle GS can still just have a sword in logic, then *this* is perfectly fine.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: I believe it is the same webbing as the Back Room; but that's why I put in the comment.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: Yes, you appear behind the web. The only place I am not sure about where you spawn from a door is Jabu-Jabu MQ Basement Side Room > Basement (the red tentacle looks to be blocking going to side room, but I don't know if it blocks going to basement main).

packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,282 @@
...
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Room Before Compass Room": "true"
+    "MQ Deku Tree Compass Room Exit Door": "event(DEKU_MQ_BEFORE_COMPASS_EYE_SWITCH)"
+"MQ Deku Tree Compass Room Exit Door":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Compass Room Entry Door": "true"
+    "MQ Deku Tree MQ Compass Room Entrance": "true"
+"MQ Deku Tree MQ Compass Room Entrance":
```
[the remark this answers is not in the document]
Joshua8600: Yes there is. Action shuffle. If you can't climb, you can go down or use hook anywhere.

packages/data/src/world/mq/deku_tree_mq.yml
XenoWars: Still true. how are you gonna kill the spider if you can't shoot through the crate

packages/data/src/world/mq/deku_tree_mq.yml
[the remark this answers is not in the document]
XenoWars: doesnt matter. there is still no other entrance to this room. All the logic can just be done via one region, and there is no need to return to the door due to being able to savewarp and walk back.

packages/data/src/world/mq/deku_tree_mq.yml
XenoWars: Still, this was also decided by AmazingAmpharos for OOTMM, who was also the main logic writer for OOTR.

packages/data/src/world/mq/deku_tree_mq.yml
XenoWars: and it can just be changed then. the has_fire is completely redundant right now, and who knows if the switch rando will even still use the events since you'd need to dynamically change the requirements if it becomes a hold switch or hammer switch.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: Fixed typo and added hovers.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: Added webs event; bow alone works for child.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: Moved the note.

Sticks is in logic still from the event in the basement room; the events are the same name.

And direction does matter as the Door to Water Room is barred when coming from Water Room; whereas door to Basement is open when clearing the room.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: I added a note for hookshot, as I can't seem to replicate it. But I did do it before when I wrote the logic.

packages/data/src/world/mq/deku_tree_mq.yml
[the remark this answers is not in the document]
XenoWars: The logic for the grass and lower chest could in the future just be:
`(can_climb || can_jump || hookshot_anywhere) && can_cut_grass` with the grass macro obvoiusly just for the grass. The upper chest would be `climb_anywhere || (can_climb && (has_bombchu || has_mask_blast || (can_play_time && has_explosives_or_hammer)))`
Not very complicated at all.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: But why delete it if the "true" is already technically there? Why delete it just to re-add it if we need it?

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: This isn't OOTR. Saying "OOTR does it" has not been a valid argument for over a year, Nax even said that.

XenoWars: That isn't even the point. the point is that both Ampharos and I agreed it should be a mercy.

Joshua8600: I guess we could make it a trick.

But there's a huge amount of difference in difficulties when we use mercies or not. It's inconsistent.

packages/data/src/world/mq/deku_tree_mq.yml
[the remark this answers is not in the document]
Joshua8600: I added your thing from discord at the end, commented out, for future reference.

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: Trick added.

packages/data/src/world/mq/deku_tree_mq.yml
```diff
@@ -2,126 +2,286 @@
...
+  exits:
+    "MQ Deku Tree Room Before Water Room Door to Water Room": "true"
+    "MQ Deku Tree Water Room Front": "true"
+"MQ Deku Tree Water Room Front":
+  dungeon: DT
+  exits:
+    "MQ Deku Tree Water Room Door to Room Before Water Room": "true"
+    "MQ Deku Tree Water Room Back": "(is_child && has_shield) || can_longshot || (can_hookshot && has_iron_boots) || trick(OOT_DEKU_MQ_WATER_ROOM_SPIKE_ROLL)"
+  events:
+    MQ_DEKU_WATER_TORCHES: "can_use_sticks || (can_use_bow && can_play_elegy)" #You can do this as both adult and child, without a shield, by rolling under the spike.
```
XenoWars, line 185:
```suggestion
   "MQ_DEKU_WATER_TORCHES: "(is_child && has_shield && can_use_sticks) || (can_use_bow && can_play_elegy) || (can_use_sticks && trick(OOT_DEKU_MQ_WATER_ROOM_SPIKE_ROLL))"
```
Also, only adult can longshot over the log. Not sure about the bow over the log though.

XenoWars (changes requested):

packages/data/src/world/mq/deku_tree_mq.yml
Joshua8600: Fixed.

Just tested; child can longshot just fine over it. Child and bow+elegy works too, just is a little precise (just tested).

XenoWars (approved):

PR #758, opened 2024-07-18
Dodongo MQ Rework

Joshua8600:


packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
+    MQ_DC_OPEN_SKULL: "(hookshot_anywhere || climb_anywhere) && has_explosives"
   locations:
-    "MQ Dodongo Cavern Map Chest": "has_bombflowers || can_hammer || has_blue_fire_arrows_mudwall"
+    "MQ Dodongo Cavern Map Chest": "has_bombflowers || can_hammer || has_blue_fire || has_blue_fire_arrows_mudwall"
     "MQ Dodongo Cavern Lobby Scrub Front": "business_scrub(0x1c)"
     "MQ Dodongo Cavern Lobby Scrub Back": "business_scrub(0x1d)"
   gossip:
-    "Dodongo Cavern Gossip": "has_bombflowers || can_hammer || has_blue_fire_arrows_mudwall"
-"Dodongo Cavern Staircase":
+    "Dodongo Cavern Gossip": "has_bombflowers || can_hammer || has_blue_fire || has_blue_fire_arrows_mudwall"
```
XenoWars, line 24: remove blue fire

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
+    MQ_DC_STAIRCASE_SWITCH: "true"
+    DC_MQ_SHORTCUT: "has_explosives_or_hammer || can_use_din"
+"Dodongo Cavern Staircase Room Lower":
   dungeon: DC
   exits:
-    "Dodongo Cavern Main": "true"
-    "Dodongo Cavern Upper Staircase": "has_bombflowers"
+    "Dodongo Cavern Main Room Lower": "true"
+    "Dodongo Cavern Staircase Room Upper": "event(DC_MQ_STAIRCASE) || climb_anywhere || hookshot_anywhere"
+    "Dodongo Cavern MQ Song of Time Block Room": "has_bombflowers || can_hammer || has_blue_fire || has_blue_fire_arrows_mudwall"
```
XenoWars, line 40: Remove Blue Fire

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
   events:
-    STICKS: "soul_deku_baba && ((has_bombflowers || can_hammer) && (has_weapon || can_boomerang))"
+    STICKS: "soul_deku_baba && ((has_bombflowers || can_hammer || has_blue_fire || has_blue_fire_arrows_mudwall) && (has_weapon || can_boomerang))"
```
XenoWars, line 42: Remove Blue Fire

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
+"Dodongo Cavern Vanilla Bomb Bag Room Upper":
   dungeon: DC
   exits:
+    "Dodongo Cavern Vanilla Bomb Bag Room Lower": "true"
+    "Dodongo Cavern Main Room Upper Skull Side": "event(DC_MQ_SHORTCUT) || climb_anywhere"
     "Dodongo Cavern Room After Upper Lizalfos": "true"
-    "Dodongo Cavern Torch Room": "true"
   events:
-    "DC_MQ_SHORTCUT": "true"
+    DC_MQ_SHORTCUT: "has_bombflowers || has_fire || has_explosives_or_hammer"
```
XenoWars, line 107: Instead of "has_fire", it should be "can_use_din || can_use_bow" as you do not need fire arrows specifically.

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
+    "Dodongo Cavern Vanilla Bomb Bag Room Lower": "event(MQ_DC_CLEAR_LARVAE_ROOM)"
+  events:
+    MQ_DC_CLEAR_LARVAE_ROOM: "soul_enemy(SOUL_ENEMY_GOHMA_LARVA) && can_damage"
+  locations:
+    "MQ Dodongo Cavern Larvae Room Chest": "event(MQ_DC_CLEAR_LARVAE_ROOM)"
+    "MQ Dodongo Cavern GS Larve Room": "gs && can_damage_skull"
+"Dodongo Cavern Room Before Upper Lizalfos":
+  dungeon: DC
+  exits:
+    "Dodongo Cavern Upper Lizalfos": "(can_use_sticks && has_goron_bracelet) || has_explosives_or_hammer || has_blue_fire || has_blue_fire_arrows_mudwall" #Needs macro for getting past enemies still in path
```
XenoWars, line 124: Remove blue fire. Also, you still need to burn some webs to get to the end.

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
+"Dodongo Cavern Upper Lizalfos":
+  dungeon: DC
+  exits:
+    "Dodongo Cavern Room Before Upper Lizalfos": "event(MQ_DC_UPPER_LIZALFOS_CLEAR)"
+    "Dodongo Cavern Room After Upper Lizalfos": "event(MQ_DC_UPPER_LIZALFOS_CLEAR)"
+    #"Dodongo Cavern Lower Lizalfos": "has_explosives_or_hammer || hookshot_anywhere || climb_anywhere"
+  events:
+    MQ_DC_UPPER_LIZALFOS_CLEAR: "soul_lizalfos_dinalfos && (can_use_sticks || has_weapon || can_use_slingshot)"
+  locations:
+    "MQ Dodongo Cavern GS Upper Lizalfos": "gs && (has_explosives_or_hammer || longshot_anywhere)"
```
XenoWars, line 148: Should probably include the climb anywhere method for this, assuming you can climb up the pillar after climbing over the rocks.

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
+    "MQ Dodongo Cavern Pot Miniboss 3": "true"
+    "MQ Dodongo Cavern Pot Miniboss 4": "true"
+    "MQ Dodongo Cavern Heart Lizalfos Room": "has_explosives_or_hammer || hookshot_anywhere || climb_anywhere" #In Lower section, putting it here too because of the lizalfos event won't trigger when starting in opposite section.
 "Dodongo Cavern Lower Tunnel":
   dungeon: DC
   exits:
-    "Dodongo Cavern Lower Lizalfos": "can_use_bow || ((has_bombflowers || can_use_din) && can_use_slingshot)"
+    "Dodongo Cavern Corridor Before Lower Lizalfos": "event(MQ_DC_LOWER_TUNNEL_EYE_SWITCH)"
+    "Dodongo Cavern Main Room Lower": "true"
+    "Dodongo Cavern Lower Tunnel Side Room": "has_bombflowers || has_explosives_or_hammer || has_blue_fire || has_blue_fire_arrows_mudwall"
```
XenoWars, line 159: Remove blue fire

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
-    "Dodongo Cavern Lower Tunnel": "soul_lizalfos_dinalfos && (can_use_sticks || has_weapon || can_use_slingshot)"
-    "Dodongo Cavern Poe Room": "soul_lizalfos_dinalfos && (can_use_sticks || has_weapon || can_use_slingshot)"
+    "Dodongo Cavern Corridor Before Lower Lizalfos": "event(MQ_DC_LOWER_LIZALFOS_CLEAR)"
+    "Dodongo Cavern Poe Room": "event(MQ_DC_LOWER_LIZALFOS_CLEAR)"
+    #"Dodongo Cavern Upper Lizalfos": "climb_anywhere || longshot_anywhere"
+  events:
+    MQ_DC_LOWER_LIZALFOS_CLEAR: "soul_lizalfos_dinalfos && (can_use_sticks || has_weapon || can_use_slingshot)"
   locations:
     "MQ Dodongo Cavern Heart Lizalfos Room": "true"
+    "MQ Dodongo Cavern GS Upper Lizalfos": "gs && (climb_anywhere || longshot_anywhere) && (has_explosives_or_hammer || longshot_anywhere)" #In Upper section, putting it here too because of the lizalfos event won't trigger when starting in opposite section.
```
XenoWars, line 188: Should be able to just climb up the pillar right? in which case you can just have the "can_damage_skull" macro for climb anywhere. No real reason to list longshot anywhere twice either.

"gs && (longshot_anywhere || (climb_anywhere && can_damage_skull))"

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
-    "MQ Dodongo Cavern GS Near Boss": "gs"
     "MQ Dodongo Cavern Pot Before Boss 1": "true"
     "MQ Dodongo Cavern Pot Before Boss 2": "true"
+"Dodongo Cavern Boss Loop Start":
+  dungeon: DC
+  exits:
+    "Dodongo Cavern Pre-Boss Lobby": "true"
+    "Dodongo Cavern Boss Loop After Fire Wall": "is_adult || climb_anywhere || hookshot_anywhere || event(MQ_DC_BOSS_LOOP_FIRE_WALL_CRYSTAL_BEFORE_FIRE)"
+  events:
+    MQ_DC_BOSS_LOOP_FIRE_WALL_CRYSTAL_BEFORE_FIRE: "has_bombs || has_bombchu"
```
XenoWars, line 242: this shouldnt be an event since it's temporary, and can just be has_explosives.

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
-    "MQ Dodongo Cavern Pot Before Boss Loop 3": "true"
-    "MQ Dodongo Cavern Pot Before Boss Loop 4": "true"
+"Dodongo Cavern Boss Loop After Fire Wall":
+  dungeon: DC
+  exits:
+    "Dodongo Cavern Boss Loop Side Room": "true"
+    "Dodongo Cavern Boss Loop Start": "event(MQ_DC_BOSS_LOOP_FIRE_WALL_CRYSTAL_AFTER_FIRE) || climb_anywhere || hookshot_anywhere"
+    "Dodongo Cavern Boss Loop After Armos Wall": "climb_anywhere || hookshot_anywhere || has_fire || has_explosives"
+  events:
+    MQ_DC_BOSS_LOOP_FIRE_WALL_CRYSTAL_AFTER_FIRE: "has_weapon || has_bombs || can_use_mask_blast || has_bombchu || can_use_bow || can_use_slingshot || can_hookshot || can_use_sticks || can_hammer || can_boomerang" #Many ways to hit crystal switch
```
XenoWars, line 253: Seems like this could really use that crystal switch macro

XenoWars (changes requested):

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
   dungeon: DC
   exits:
     "Death Mountain": "true"
-    "Dodongo Cavern Main": "has_bombflowers || can_hammer || has_blue_fire_arrows_mudwall"
-"Dodongo Cavern Main":
+    "Dodongo Cavern Main Room Lower": "has_bombflowers || can_hammer || has_blue_fire || has_blue_fire_arrows_mudwall"
```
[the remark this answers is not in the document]
Joshua8600: Done.

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
-    "Dodongo Cavern Staircase": "has_bombflowers || can_hammer"
-    "Dodongo Cavern Upper Ledges": "has_explosives_or_hammer || can_use_din"
-    "Dodongo Cavern Lower Tunnel": "has_explosives_or_hammer || (event(DC_MQ_SHORTCUT) && has_goron_bracelet)"
-    "Dodongo Cavern Bomb Bag Ledge": "is_adult"
+    "Dodongo Cavern": "has_explosives_or_hammer || has_blue_fire || has_blue_fire_arrows_mudwall" #We need to see if you can go backwards without breaking it, like other mudwalls.
+    "Dodongo Cavern Main Room Inside Skull": "event(MQ_DC_OPEN_SKULL)"
+    "Dodongo Cavern Staircase Room Lower": "event(MQ_DC_STAIRCASE_SWITCH)"
+    "Dodongo Cavern Main Room Upper Skull Side": "event(MQ_DC_PILLAR_RAISE) || climb_anywhere || hookshot_anywhere" #You can use hook anywhere to get on top of the skull and again to get on the bridge without longshot.
+    "Dodongo Cavern Main Room Upper Entrance Side": "climb_anywhere || longshot_anywhere"
+    "Dodongo Cavern Lower Tunnel": "has_explosives_or_hammer || (event(DC_MQ_SHORTCUT) && has_goron_bracelet) || has_blue_fire || has_blue_fire_arrows_mudwall"
```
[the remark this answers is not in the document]
Joshua8600: Done.

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
+    "Dodongo Cavern Bomb Bag Ledge": "is_adult || climb_anywhere || can_hookshot"
+  events:
+    MQ_DC_PILLAR_RAISE: "has_bombflowers || can_hammer"
+    MQ_DC_OPEN_SKULL: "(hookshot_anywhere || climb_anywhere) && has_explosives"
   locations:
-    "MQ Dodongo Cavern Map Chest": "has_bombflowers || can_hammer || has_blue_fire_arrows_mudwall"
+    "MQ Dodongo Cavern Map Chest": "has_bombflowers || can_hammer || has_blue_fire || has_blue_fire_arrows_mudwall"
```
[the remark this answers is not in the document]
Joshua8600: Done.

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: Done.

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: You cannot do that.

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: We're changing the has_explosives later on to a ranged macro. Also, this event is only used right here so I don't think it makes much difference.

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: Changed.

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: I agree. xD

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -2,159 +2,279 @@
...
+"Dodongo Cavern Main Room Lower":
   dungeon: DC
   exits:
-    "Dodongo Cavern": "true"
-    "Dodongo Cavern Skull": "has_explosives"
-    "Dodongo Cavern Staircase": "has_bombflowers || can_hammer"
-    "Dodongo Cavern Upper Ledges": "has_explosives_or_hammer || can_use_din"
-    "Dodongo Cavern Lower Tunnel": "has_explosives_or_hammer || (event(DC_MQ_SHORTCUT) && has_goron_bracelet)"
-    "Dodongo Cavern Bomb Bag Ledge": "is_adult"
+    "Dodongo Cavern": "has_explosives_or_hammer || has_blue_fire || has_blue_fire_arrows_mudwall" #We need to see if you can go backwards without breaking it, like other mudwalls.
```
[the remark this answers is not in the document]
XenoWars: You actually can by using the front platform, have to throw it before it explodes. but this may become moot anyway if you can walk through it. Which should be testable on the OOTR branches that have boss mixed pools right?

packages/data/src/world/mq/dodongo_cavern_mq.yml
XenoWars: Yes you can. Tested it before i did the time travel.

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -144,11 +144,12 @@
...
   exits:
-    "Dodongo Cavern Room Before Upper Lizalfos": "event(MQ_DC_UPPER_LIZALFOS_CLEAR)"
-    "Dodongo Cavern Room After Upper Lizalfos": "event(MQ_DC_UPPER_LIZALFOS_CLEAR)"
+    "Dodongo Cavern Room Before Upper Lizalfos": "event(MQ_DC_UPPER_LIZALFOS_CLEAR_FROM_UPPER) || (event(MQ_DC_UPPER_LIZALFOS_CLEAR_FROM_LOWER)"
+    "Dodongo Cavern Room After Upper Lizalfos": "event(MQ_DC_UPPER_LIZALFOS_CLEAR_FROM_UPPER) || (event(MQ_DC_UPPER_LIZALFOS_CLEAR_FROM_LOWER)"
     #"Dodongo Cavern Lower Lizalfos": "has_explosives_or_hammer || hookshot_anywhere || climb_anywhere"
   events:
-    MQ_DC_UPPER_LIZALFOS_CLEAR: "soul_lizalfos_dinalfos && (can_use_sticks || has_weapon || can_use_slingshot)"
+    MMQ_DC_UPPER_LIZALFOS_CLEAR_FROM_UPPER: "soul_lizalfos_dinalfos && (can_use_sticks || has_weapon || can_use_slingshot)"
+    MQ_DC_LOWER_LIZALFOS_CLEAR_FROM_UPPER: "(has_explosives_or_hammer || hookshot_anywhere || climb_anywhere) && soul_lizalfos_dinalfos && (can_use_sticks || has_weapon || can_use_slingshot)"
```
XenoWars, line 152: Forgot the time travel

XenoWars (changes requested):

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: Then I don't know what you mean. What line are you even referring to?

XenoWars: This one right here. GS Upper Lizalfos. You can climb around the rocks, drop down, then climb up the pillar.

Joshua8600: But that's going from Upper>Lower>Upper GS, not Upper>GS; thats why the Upper GS is listed in the lower region too

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: Fixed.

packages/data/src/world/mq/dodongo_cavern_mq.yml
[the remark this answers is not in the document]
Joshua8600: I tried, but it didn't look like it reached. Were you adult or child?

packages/data/src/world/mq/dodongo_cavern_mq.yml
XenoWars: But there is no exit between the regions here due to the lizalfos fights. So need to account for it here as well. The way i did it in vanilla was putting the time travel requirement on the exits (because there are not yet any fight events). Here, you could require the lower fight events to go the lower region, and vice versa. But you would still need this climb anywhere method in the upper region cause you may not have song of time.

Joshua8600: That's what I did, though? And what does that have to do with the GS?

packages/data/src/world/mq/dodongo_cavern_mq.yml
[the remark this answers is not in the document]
XenoWars: adult. i can try as child

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: tell me what to put for the webs. i cant picture it

packages/data/src/world/mq/dodongo_cavern_mq.yml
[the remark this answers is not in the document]
XenoWars: yeah adult can do it easily. child doesnt appear to be fast enough to get even the throw close enough.

packages/data/src/world/mq/dodongo_cavern_mq.yml
XenoWars: No, that has not been done yet. There are currently no exits going between the lower and upper lizalfos fights. And it has to do with the GS because you may not have SoT to go between the two regions logically.

Joshua8600: I do have the exits there, just commented out because of the lizalfos thing. now that the time travel is in, could they be implemented and then it would work just fine? Lines 149 and 189

packages/data/src/world/mq/dodongo_cavern_mq.yml
XenoWars: ah. no webs, but for the sticks you would need to be able to get the torch lit. So probably an event for the torch on the other side (requires lighting the two torches on the front side).

packages/data/src/world/mq/dodongo_cavern_mq.yml
XenoWars: They can be implemented, but they would need to require the relevant events to go through the doors in their respective region. Cause otherwise you could exit to them and be expected to go through the doors when you dont have Song of Time due to having access to the event that does not require SoT. And because they would need to require the events, since you may not be able to do the events from where you entered, you need to be able to account for that with the locations in these regions (already done for everything but climb anywhere for this GS from upper).

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -151,7 +153,7 @@
     MQ_DC_UPPER_LIZALFOS_CLEAR_FROM_UPPER: "soul_lizalfos_dinalfos && (can_use_sticks || has_weapon || can_use_slingshot)"
     MQ_DC_LOWER_LIZALFOS_CLEAR_FROM_UPPER: "(time_travel_at_will && (has_explosives_or_hammer || hookshot_anywhere || climb_anywhere)) && soul_lizalfos_dinalfos && (can_use_sticks || has_weapon || can_use_slingshot)"
   locations:
-    "MQ Dodongo Cavern GS Upper Lizalfos": "gs && (has_explosives_or_hammer || longshot_anywhere)"
+    "MQ Dodongo Cavern GS Upper Lizalfos": "gs && (has_explosives_or_hammer || climb_anywhere || longshot_anywhere)"
```
XenoWars, line 156: climbing will need some way to damage.

XenoWars (changes requested):

packages/data/src/world/mq/dodongo_cavern_mq.yml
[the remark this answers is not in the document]
Joshua8600: Added adult with bombflowers

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: Done.

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: Fixed.

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: Okay added climb_anywhere to the GS from Upper

packages/data/src/world/mq/dodongo_cavern_mq.yml
```diff
@@ -93,7 +93,7 @@
   dungeon: DC
   exits:
     "Dodongo Cavern Main Room Upper Entrance Side": "true"
-    "Dodongo Cavern Room Before Upper Lizalfos": "can_use_sticks || has_fire"
+    "Dodongo Cavern Room Before Upper Lizalfos": "(can_use_sticks && event(OPEN_MQ_DC_LARVAE_ROOM)) || has_fire"
```
XenoWars, line 96: This can actually just be true right now, since the transition occurs before the web.

XenoWars (changes requested):

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: is this wrong though? if its right, I'd rather just leave it

XenoWars: It's fine. Just would need changed if plane transitions are ever shuffled.

packages/data/src/world/mq/dodongo_cavern_mq.yml
Joshua8600: added "can_damage_skull"

XenoWars (approved):

PR #760, opened 2024-07-18
Ganon Tower Rework

Joshua8600:


XenoWars (changes requested):
Well, i was going to joke and say it's all wrong since it's Ganon's Tower; it's simple. But you're actually missing three rooms. The ones between each fight and the boss door.

XenoWars (approved):

PR #764, opened 2024-07-26
Separate Pirate Fortress ER Option

XenoWars:
This adds an option that separates the Pirate Fortress entrances from the relevant shuffles. The Sewers exit door will be shuffled if this and Extra Interiors are enabled, the Sewers gate and two interior entrances will be shuffled if this and Overworld ER are enabled, and if the previous is true then Pirate Fortress will no longer be considered a dungeon and shuffle the main entrance among the overworld as well.


packages/core/lib/combo/patch-build/randomizer.ts
```diff
@@ -720,7 +720,7 @@ const gameEntrances = (worldId: number, game: Game, logic: LogicResult) => {
   return padBuffer16(toU32Buffer(data));
 };
 
-const randomizerDungeonsBits = (worldId: number, logic: LogicResult): Buffer => {
+const randomizerDungeonsBits = (worldId: number, logic: LogicResult, settings: Settings): Buffer => {
```
CelestialKitsune, line 723: No need to pass the `settings`, they're accessible through the `logic` object, please use `logic.settings` instead within the function

packages/core/lib/combo/patch-build/randomizer.ts
```diff
@@ -761,6 +761,9 @@ const randomizerDungeonsBits = (worldId: number, logic: LogicResult): Buffer =>
 
   for (let i = 0; i < DUNGEONS_PRECOMPLETED.length; ++i) {
     const dungeon = DUNGEONS_PRECOMPLETED[i];
+    if (settings.erPiratesWorld && world.preCompleted.has('PF')) {
```
CelestialKitsune, line 764: This will never be true as PF cannot be a pre-completed dungeon.

CelestialKitsune (changes requested):

packages/core/lib/combo/patch-build/randomizer.ts
XenoWars: Well, PF is part of the pre-completed dungeon list, so I'm not quite sure what you mean by that.

CelestialKitsune: Check the [`selectPreCompletedDungeons`](https://github.com/OoTMM/OoTMM/blob/master/packages/core/lib/combo/logic/solve.ts#L810-L818) function and the other functions it calls. Only Major Dungeons can be pre-completed.

PR #781, opened 2024-09-12
Erase FlashRAM sector before write when necessary

Polprzewodnikowy:
This PR implements correct sequence of operations when accessing FlashRAM for write operations. `writeFlash` function was rewritten to use sector buffer (16 kiB) allocated on the heap for read-modify-write operations. Previous solution relied on inaccurate implementations of the FlashRAM chip: data was simply written to the save memory without erasing it before doing this operation.

From my knowledge OoTMM is very memory constrained so testing is needed if the heap allocation could fail in some circumstances. My limited testing didn't show any problems so far, but I don't know this project internals very much so I might miss something important.


networkfusion:
I just wanted to note that this PR fixes the issue that prevents the ROM from running on flashcarts and emulators (such as the SummerCart64) that properly handle the FlashRAM hardware properly.

PR #784, opened 2024-10-27
MM Megaton Hammer

Aegiker:
Adds Megaton Hammer as an item and extends support for more custom items


packages/core/src/mm/actors/Player.c
```diff
@@ -933,6 +934,88 @@ Gfx gLinkChildGoronBraceletDL[] = {
...
 
+/* Conveniently, OoT's Megaton Hammer only used GPK textures that still have equivalents in MM */
+const u32 gDekuStickTex = 0x0400CC30;
+const u32 gHilite1Tex = 0x0400C830;
+
+const Vtx sHammerVtx[] = {
+#include "sHammerVtx.vtx.inc"
+};
+
+Gfx gHammerDL[] = {
```
CelestialKitsune, line 945:
```suggestion
static Gfx gHammerDL[] = {
```

CelestialKitsune (changes requested):
It all seems fine to me, besides that one suggested change.

packages/core/src/mm/actors/Player.c
Aegiker: oops

packages/core/src/mm/actors/Player.c
```diff
@@ -1644,3 +1738,437 @@ void Player_UseItem(GameState_Play* play, Actor_Player* this, s16 itemId)
...
+    /* PLAYER_MWA_HAMMER_FORWARD */
+    //{ &gPlayerAnim_link_hammer_hit, &gPlayerAnim_link_hammer_hit_end, &gPlayerAnim_link_hammer_hit_endR, 3, 10 },
+    { (PlayerAnimationHeader*)0x0400E4E8, (PlayerAnimationHeader*)0x0400E4F0, (PlayerAnimationHeader*)0x0400E4E0, 3, 10 },
+    /* PLAYER_MWA_HAMMER_SIDE */
+    //{ &gPlayerAnim_link_hammer_side_hit, &gPlayerAnim_link_hammer_side_hit_end, &gPlayerAnim_link_hammer_side_hit_endR, 2, 11 },
+    { (PlayerAnimationHeader*)0x0400E4D8, (PlayerAnimationHeader*)0x0400E4D0, (PlayerAnimationHeader*)0x0400E4C8, 2, 11 },
+};
+
+AttackAnimInfo* Player_GetMeleeAttackAnimInfo(void* a0, Actor_Player* player, PlayerMeleeWeaponAnimation meleeWeaponAnim) {
+    asm("sw $a1, 0x0018 ($sp)"); /* keep player in a1 */
```
ZoeyZolotova, line 1857: Are we okay with this? Is there no way around this or is it fine as is? I suppose it's not much different then having an asm hook right? Like
```
Player_GetMeleeAttackAnimInfoHook:
  addiu sp, sp, -0x20
  // store ra
  jal Player_GetMeleeAttackAnimInfo
    // store a1
  // load a1
  // load ra
  jr ra
  addiu sp, sp, 0x20
```

ZoeyZolotova: Though this way does assume that 0x0018 (sp) inside the code of this function is available.

packages/core/include/combo/data/items.h
```diff
@@ -306,6 +306,9 @@
 #define ITEM_MM_BOOTS_HOVER                 0xb2
 #define ITEM_MM_TUNIC_GORON                 0xb3
 #define ITEM_MM_TUNIC_ZORA                  0xb4
+#define ITEM_MM_HAMMER                      0xb5
+#define ITEM_MM_CUSTOM_MAX                  0xb6
+
 /* If we want to use values above 0xb5 we need to fix a lot of code related to magic arrows */
```
ZoeyZolotova, line 312: Can this comment be updated to reflect the new safest upper value to use?

packages/core/src/mm/actors/Player.S
```diff
@@ -1224,3 +1224,7817 @@ PATCH_END
 PATCH_START 0x8085e958
 .int 0
 PATCH_END
+
+/* Hammer Related ASM */
+
+PATCH_START 0x80123960
```
ZoeyZolotova, line 1230: I believe this patch can be removed and replaced with `PATCH_FUNC(0x80123960, Player_CustomActionToModelGroup)`

packages/core/src/mm/actors/Player.S
```diff
@@ -1224,3 +1224,7817 @@ PATCH_END
...
+  addiu     sp, sp, -0x20
+  sw        ra, 0x0014 (sp)
+  jal Player_CustomActionToModelGroup
+  nop
+  lw        ra, 0x0014 (sp)
+  jr        ra
+  addiu     sp, sp, 0x20
+PATCH_END
+
+PATCH_START 0x80128DDC
```
ZoeyZolotova, line 1240: There appears to be a branch instruction at 0x801289d8 that branches to the second line in this patch. Could that cause any issues since it skips setting `a0`?

packages/core/src/mm/actors/Player.S
```diff
@@ -1224,3 +1224,7817 @@ PATCH_END
...
+  addiu     a0, s2, 0x0
+  jal       Player_GetCustomSwordLength
+  addiu     a1, s1, 0x0
+  j         0x80128E24
```
ZoeyZolotova, line 1244: I don't know about the others, but my personal preference for this would be to add `nop` 15 times here since that code becomes unreachable.

packages/core/src/mm/actors/Player.S
```diff
@@ -1224,3 +1224,7817 @@ PATCH_END
...
+  sw        ra, 0x0014 (sp)
+  jal Player_CustomActionToModelGroup
+  nop
+  lw        ra, 0x0014 (sp)
+  jr        ra
+  addiu     sp, sp, 0x20
+PATCH_END
+
+PATCH_START 0x80128DDC
+  addiu     a0, s2, 0x0
```
ZoeyZolotova, line 1241: Doesn't bother me much, but for readability I think this would be better as `move a0, s2`. Same in other similar instructions.

packages/core/src/mm/actors/Player.c
```diff
@@ -1644,3 +1738,437 @@ void Player_UseItem(GameState_Play* play, Actor_Player* this, s16 itemId)
...
+
+/* Hammer Stuff */
+
+s32 Player_CustomActionToModelGroup(Actor_Player* player, s32 itemAction) {
+    u8* sActionModelGroups = (u8*)0x801BFF3C; /* using original table also means original glitches, if that matters */
+    s32 modelGroup = sActionModelGroups[itemAction];
+
+    if (itemAction == PLAYER_CUSTOM_IA_HAMMER) {
+        modelGroup = sActionModelGroups[7]; /* uses deku stick model group but does not draw deku stick because of the way the original draw code for it works */
+    }
```
ZoeyZolotova, line 1750: Maybe this would read better as 
```suggestion
    if (itemAction == PLAYER_CUSTOM_IA_HAMMER) {
        itemAction = PLAYER_IA_DEKU_STICK;
    }
    s32 modelGroup = sActionModelGroups[itemAction];
```
?

packages/core/src/mm/actors/Player.c
```diff
@@ -1644,3 +1738,437 @@ void Player_UseItem(GameState_Play* play, Actor_Player* this, s16 itemId)
...
+*/
+void Player_SetCustomItemActionUpperFunc(GameState_Play* play, Actor_Player* player) {
+    PlayerUpperActionFunc* sPlayerUpperActionUpdateFuncs = (PlayerUpperActionFunc*)OverlayAddr(0x8085c9f0);
+    void (*Player_SetUpperAction)(GameState_Play* play, Actor_Player* this, PlayerUpperActionFunc upperActionFunc) = OverlayAddr(0x8082f43c);
+
+    if (player->heldItemAction == PLAYER_CUSTOM_IA_HAMMER) {
+        Player_SetUpperAction(play, player, sPlayerUpperActionUpdateFuncs[6]);
+    } else {
+        Player_SetUpperAction(play, player, sPlayerUpperActionUpdateFuncs[player->heldItemAction]);
+    }
```
ZoeyZolotova, line 1771: Maybe it's just my style preference but i'd write it more like this.
```suggestion
    s8 heldItemAction = player->heldItemAction;
    if (heldItemAction == PLAYER_CUSTOM_IA_HAMMER) {
        heldItemAction = PLAYER_IA_SWORD_TWO_HANDED;
    }
    Player_SetUpperAction(play, player, sPlayerUpperActionUpdateFuncs[heldItemAction ]);
```

packages/core/src/mm/actors/Player.c
```diff
@@ -1644,3 +1738,437 @@ void Player_UseItem(GameState_Play* play, Actor_Player* this, s16 itemId)
...
+}
+
+void Player_RunCustomItemActionInitFunc(GameState_Play* play, Actor_Player* player, s32 itemAction) {
+    PlayerInitItemActionFunc* sPlayerItemActionInitFuncs = (PlayerInitItemActionFunc*)OverlayAddr(0x8085cb3c);
+    
+    if (player->heldItemAction == PLAYER_CUSTOM_IA_HAMMER) {
+        sPlayerItemActionInitFuncs[6](play, player);
+    } else {
+        sPlayerItemActionInitFuncs[itemAction](play, player);
+    }
```
ZoeyZolotova, line 1781: Similar as above.

packages/core/src/mm/objects/gameplay_keep.S
```diff
@@ -0,0 +1,33 @@
+#include <combo.h>
+
+/* Hammer Animations - Patch AnimHeader Frame Count */
+
+/* gPlayerAnim_Link_otituku_w -> gPlayerAnim_link_hammer_hit */
```
ZoeyZolotova, line 5: I'm just curious, which unused animations did these replace?

packages/core/src/mm/ovl_patch/kaleido_scope.c
```diff
@@ -334,15 +348,16 @@ static u32 sCustomIcons[] = {
...
 
-s8 gPlayerFormCustomItemRestrictions[5][8] =
+s8 gPlayerFormCustomItemRestrictions[5][ITEM_MM_CUSTOM_MAX - ITEM_MM_CUSTOM_MIN] =
 {
     { 0, 0, 0, 0, 0, 0, 0, 0 },
     { 0, 0, 0, 0, 0, 0, 0, 0 },
     { 0, 0, 0, 0, 0, 0, 0, 0 },
     { 0, 0, 0, 0, 0, 0, 0, 0 },
-    { 1, 1, 1, -1, -1, -1, -1, 0 },
+    { 1, 1, 1, -1, -1, -1, -1, -1 },
```
ZoeyZolotova, line 360: `-1` means it'll be forced to be enabled (such as in situations like being underwater or indoors), that's why it's used for boots and tunics. If you set it to `1` it should match the behavior of other items.

packages/core/src/mm/actors/Player.S
```diff
@@ -1224,3 +1224,7817 @@ PATCH_END
...
+  j Player_BeginHammerShockwaveCheck
+  nop
+PATCH_END 
+
+/* Hammer link_animetion Patches */
+/* These animations should overwrite unused Ganon/Ganondorf related animations from OoT */
+/* PATCH_VROM link_animation start + frame data offset */
+
+/* gPlayerAnim_Link_otituku_w -> gPlayerAnim_link_hammer_hit */
+PATCH_VROM 0x65d000 + 0xA45D0
```
ZoeyZolotova, line 1314: @Nax I hope you have a suggestion on how this can be improved.

packages/core/src/mm/actors/Player.c
Aegiker: an asm hook could be used instead yes, I think it's okay though. I'm kind of attached to it now though so I'm biased. I stepped through this function in a debugger to see how the compiler wrote it before and after making this.

packages/core/include/combo/data/items.h
Aegiker: I will update it. For reference, the safest upper value is now ITEM_MM_CUSTOM_MAX

packages/core/src/mm/actors/Player.S
Aegiker: It does not matter because the function in question does not actually use that argument for anything. I don't remember why I set a0 at all but replacing it with a nop seemed to work ok

packages/core/src/mm/actors/Player.S
Aegiker: Thank you for pointing this out. That patch was written before I found out that PATCH_FUNC existed :p

packages/core/src/mm/actors/Player.c
Aegiker: since the index is fixed I might as well skip using the table and return the model group directly
```
    if (itemAction == PLAYER_CUSTOM_IA_HAMMER) return 10;

    u8* sActionModelGroups = (u8*)0x801BFF3C;
    s32 modelGroup = sActionModelGroups[itemAction];
      if ((modelGroup == 2) && (player->transformation == MM_PLAYER_FORM_GORON || player->transformation == MM_PLAYER_FORM_DEKU)) {
            return 1;
      }
    return modelGroup;
```

packages/core/src/mm/actors/Player.c
Aegiker: If it's alright I would prefer not to include this change because setting a variable named `heldItemAction` to something that is not actually the held item action is misleading even if the functionality is identical. I also don't know if the compiler would just optimize it out but I wonder if this change would cause the 6 to be placed in a register instead of as an added offset in the LW instruction, which I dislike. I did replace the 6 with PLAYER_IA_SWORD_HAMMER to make it clearer why this exists though. If you guys are thinking about the OoT Boomerang it would be worth considering adding a `sPlayerCustomUpperActionUpdateFuncs` alternative.

packages/core/src/mm/objects/gameplay_keep.S
Aegiker: iirc it was animations for talking to Navi/Zelda after tower escape in OoT, then turning and covering Zelda when the pile of rubble makes a noise. Also looking up at the lowering crystal containing Zelda after beating Ganondorf.

packages/core/src/mm/actors/Player.S
Aegiker: maybe these patches could be separated into a different file and `#include` could be used. I did this because there is no existing `link_animetion` file and it's not an object so there wasn't a folder for it either

packages/core/include/combo/data/items.h
ZoeyZolotova: Oh I thought that the changes in kaleido_scope.S and z_parameter.S made it automatically adjust the magic arrows issue if you increase this ITEM_MM_CUSTOM_MAX value?

Aegiker: That is correct, if you increase it the safest upper value also increases. I updated the comment to reflect that

packages/core/src/mm/actors/Player.S
ZoeyZolotova: In that case I'd say we should remove the `GameState_Play*` parameter from the function definition so that no one accidentally uses it in the future.

packages/core/include/combo/data/items.h
ZoeyZolotova: Okay that's what I thought. What I meant to ask for was what's the new safest upper limit for ITEM_MM_CUSTOM_MAX? Magic arrows use 3 slots i guess, but I assume 0xFF is not usable. Can magic arrows use 0xFE, 0xFD and 0xFC which makes 0xFB the highest safest value?

Aegiker: Oh sorry. Items greater than or equal to 0xF0 are reserved for special purposes and button actions, and kaleidoscope will attempt to avoid them. For safety I'd suggest going no higher than 0xE0, though the upper limit may be 0xED. You would have to add quite a few items to get to a point where this would be an actual problem, but I guess it is possible.

packages/core/src/mm/actors/Player.c
Nax: Yeah big no for this, very unsafe, could break at pretty much any time. Can't you just wrap the func?

packages/core/src/mm/actors/sHammerVtx.vtx.inc
```diff
@@ -0,0 +1,208 @@
+	VTX(486, 364, -100, 71, 2, 79, 0, 0, 255),
```
Nax, line 1: This shouldn't be necessary, we already extract the hammer into it's own object file for child link OoT (tho it uses segment `A`), we should reuse that.

packages/core/src/mm/actors/Player.S
Nax: Please move this into a new file (maybe `Player_hammer_anims.S` or whatever) and add a `TODO` or `FIXME` comment. This will be okay for now but eventually I want to support cross-anims so we don't have to do that.

packages/core/src/mm/actors/Player.S
Nax: I think it's fine.

packages/core/src/mm/actors/sHammerVtx.vtx.inc
Nax: Altho not urgent, I can handle that later.

packages/core/src/mm/actors/Player.c
Nax: Kinda agree with @ZoeyZolotova on this one. I guess `itemAction` would be an okay name? Much better to have a single call IMO.

Also don't worry about the compiler too much. We must aggressively optimize for space anyway, this will give it more leverage.

Nax (changes requested):
Pretty good PR and nice feature, still requires a couple more changes before we can merge

CelestialKitsune (approved):

Nax (approved):

PR #801, opened 2024-11-26
Snowhead Temple Pillar Room Logic Improvements: ground pots fix, new trick, prep for future feature

eedefeed:
In Snowhead Temple, the 'pillars' room has pillars that you can jump across. There are some freezards to kill for a chest. If you light the torches on top of the pillars, you'll unbar a door that leads goron pound switch that raises the central room pillar.

### Ground Room Pots Fix
The upper part of pillars room is not accessible without the means to melt ice, but if you go to the ground floor of the central room (where the lava pools are) you can take a door to the ground floor of the pillars room. The ground floor has pots in it. 
These pots are easily accessible without being able to melt ice, but I think the existing logic expects you to need to be able to melt ice to get these pots.

Here is a **_heavily** edited excerpt_ from the existing logic

```
"Snowhead Temple Center Level 1":
  exits:
    "Snowhead Temple Pillars Room": "can_use_fire_short_range || trick_sht_hot_water"
"Snowhead Temple Pillars Room":
  locations:
    "Snowhead Temple Pillars Room": "soul_freezard && can_kill_freezard_short_range"
    "Snowhead Temple Pot Pillars Room Upper [1..6]": "true"
    "Snowhead Temple Pot Pillars Room Lower [1..7]": "true"
"Snowhead Temple Center Level 0":
  exits:
    "Snowhead Temple Center Level 1": "true"
```
The 'Center Level 0' has no means to get to the pillars room even though there's a door to it. Center level 1 locks pillars room behind ice melting. And meanwhile the pillars room contains both the upper and lower pots. Therefore logic doesn't understand that you can get those pots without the ability to melt ice.

To fix this, I have split the pillars room into two areas, 'upper' and 'lower' and then established links between them.
The logical ways to get from lower to upper are: 1) melt ice with fire or hot water 2) short hookshot anyway 3) a new trick

You may notice that the hot water logic for entering via the upper floor door uses `trick_sht_hot_water` (maintaining the existing logic) whilst the ground floor ice uses `trick_sht_hot_water_er`. I'm not saying that `trick_sht_hot_water` is inappropriate, but I am saying that the more restrictive `trick_sht_hot_water_er` is sufficient. Outside of door rando, if you can melt the ground floor ice, then you can melt the the upper floor ice too, gaining access through that route instead. Inside of door rando, then `trick_sht_hot_water_er` will work fine.

### New trick: Climbing with hookshot
As ghunzor pointed out, you can get from the ground floor to on top of the pillars by using hookshot to kill the freezards then hookshotting to the chest that then spawns.

### Door rando
Splitting these areas needs to be done anyway because of door rando, but please don't consider these 4 areas 'complete' for door rando. In particular, the exit to central 1 from pillars upper is "true" but with door rando you would need the ability to melt ice.

### Other info
I haven't updated preset.ts since it seems hell preset handles tricks automatically now.


eedefeed:
Short hookshot anywhere to climb the pillars is pretty easy, it shouldn't need to be a trick
https://github.com/user-attachments/assets/ba3f06ce-1639-45ab-8eed-30ce2a8f89d9

packages/core/lib/combo/settings/tricks.ts
```diff
@@ -457,6 +457,11 @@ export const TRICKS: Tricks = {
     tooltip: 'Destroy the snowballs with fire arrows or bombs and then jump down.  A precise jump slash may help getting onto the platform.',
     linkVideo: 'https://youtube.com/clip/UgkxwCrliqwRiCbljfpozhmsnlzVwSUaAI0N',
   },
+  MM_SHT_PILLAR_ROOM_HOOKSHOT: {
+    game: 'mm',
+    name: 'Snowhead Temple Hookshot Up Pillar Room',
+    tooltip: 'From the ground floor of pillar room, use a hookshot to kill the freezards and then climb using the chest that spawns.',
```
Ghunzor, line 463: linkVideo: 'https://youtu.be/4WVOnjC7xdE',

Ghunzor (commented):
Gave the video link and the command for the youtube link

PR #823, opened 2025-01-17
Update gerudo_training_grounds.yml

Ghunzor:
Added swim logic to the SR not on the ground

### DIFF
--- a/packages/data/src/world/oot/gerudo_training_grounds.yml
+++ b/packages/data/src/world/oot/gerudo_training_grounds.yml
@@ -98,10 +98,10 @@
   dungeon: GTG
   locations:
     "Gerudo Training Grounds SR Water 1": "true"
-    "Gerudo Training Grounds SR Water 2": "true"
+    "Gerudo Training Grounds SR Water 2": "can_swim || can_longshot"
     "Gerudo Training Grounds SR Water 3": "true"
-    "Gerudo Training Grounds SR Water 4": "true"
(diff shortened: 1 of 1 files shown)


XenoWars (changes requested):
None of these actually need longshot, but hookshot anywhere is needed for the ceiling rupee. Getting all the rupees in one go will also require either longshot anywhere or zora tunic.

Ghunzor:
Updated

packages/data/src/world/oot/gerudo_training_grounds.yml
```diff
@@ -91,17 +91,17 @@
 "Gerudo Training Grounds Water":
   dungeon: GTG
   exits:
-    "Gerudo Training Grounds Underwater": "can_play_time && has_tunic_zora && has_iron_boots"
+    "Gerudo Training Grounds Underwater": "can_play_time && has_iron_boots && (has_tunic_zora || longshot_anywhere)"
```
XenoWars, line 94: what i meant is without swimming it hard requires the tunic or longshot anywhere to get the chest.. Also, i don't know which number it would be but only one of the silver rupees are on the ground and wouldn't need anything. The ceiling rupee still needs hookshot anywhere for it too.

I would just revert this line for now, and then i'll handle a silver rupee event since it will be easier, and I was going to do events for them all anyway.

XenoWars (changes requested):

PR #824, opened 2025-01-17
Update jabu_jabu.yml

Ghunzor:
Added B Scale logic to Raising Water Room

### DIFF
--- a/packages/data/src/world/oot/jabu_jabu.yml
+++ b/packages/data/src/world/oot/jabu_jabu.yml
@@ -86,12 +86,12 @@
   dungeon: JJ
   exits:
     "Jabu-Jabu Basement": "true"
-    "Jabu-Jabu Elevator": "true"
+    "Jabu-Jabu Elevator": "has_ranged_weapon || has_bombs" #|| Has soul_ruto && settingRutoF1=false
   locations:
     "Jabu-Jabu GS Water Switch": "gs"
-    "Jabu-Jabu Pot Alcove 1": "true"
-    "Jabu-Jabu Pot Alcove 2": "true"
-    "Jabu-Jabu Pot Alcove 3": "true"
+    "Jabu-Jabu Pot Alcove 1": "can_swim || has_hover_boots"
+    "Jabu-Jabu Pot Alcove 2": "can_swim || has_hover_boots"
+    "Jabu-Jabu Pot Alcove 3": "can_swim || has_hover_boots"
 "Jabu-Jabu Before Big Octo":
   dungeon: JJ
   exits:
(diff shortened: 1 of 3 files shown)


packages/data/src/world/oot/jabu_jabu.yml
```diff
@@ -86,12 +86,12 @@
   dungeon: JJ
   exits:
     "Jabu-Jabu Basement": "true"
-    "Jabu-Jabu Elevator": "true"
+    "Jabu-Jabu Elevator": "has_ranged_weapon || has_bombs" #|| Has soul_ruto && settingRutoF1=false
```
XenoWars, line 89: You would also need a way across the water, so either can_swim, adult, hover boots, or climb anywhere (hover boots will not work when carrying Ruto across)

packages/data/src/world/oot/jabu_jabu.yml
```diff
@@ -86,12 +86,12 @@
...
   locations:
     "Jabu-Jabu GS Water Switch": "gs"
-    "Jabu-Jabu Pot Alcove 1": "true"
-    "Jabu-Jabu Pot Alcove 2": "true"
-    "Jabu-Jabu Pot Alcove 3": "true"
+    "Jabu-Jabu Pot Alcove 1": "can_swim || has_hover_boots"
+    "Jabu-Jabu Pot Alcove 2": "can_swim || has_hover_boots"
+    "Jabu-Jabu Pot Alcove 3": "can_swim || has_hover_boots"
```
XenoWars, line 94: Climb anywhere would also work

XenoWars (changes requested):
Will also need (can_swim && is_child) || can_dive_small for the deku scrub, which should probably be given its own region.

packages/data/src/world/oot/jabu_jabu.yml
Ghunzor: That's fine since you can just toss Ruto on top of the other ledge and then hover boots across.

packages/data/src/world/oot/jabu_jabu.yml
Ghunzor: Added

packages/data/src/world/oot/jabu_jabu.yml
```diff
@@ -86,12 +86,12 @@
   dungeon: JJ
   exits:
     "Jabu-Jabu Basement": "true"
-    "Jabu-Jabu Elevator": "has_ranged_weapon || has_bombs" #|| Has soul_ruto && settingRutoF1=false
+    "Jabu-Jabu Elevator": "(can_swim || is_adult || time_travel_at_will || has_hover_boots) && (has_ranged_weapon || has_bombs)" #|| Has soul_ruto && settingRutoF1=false
```
XenoWars, line 89: time travel is not needed here, since it already can detect you can swap ages. time travel is used in other logic if you need to get a location specifically by age swapping back and forth, such as the Kak Ladder GS (age swap, climb up as adult, age swap, use din's)

XenoWars (changes requested):

PR #836, opened 2025-02-24
Forest Temple (MQ) Rework

Joshua8600:


packages/data/src/world/mq/forest_temple_mq.yml
```diff
@@ -14,137 +11,437 @@
...
-  exits:
+  exits: #Maybe make the upper branches where the event takes place its own region?
     "Forest Temple": "true"
-    "Forest Temple Antichamber": "event(FOREST_POE_4)"
-    "Forest Temple West Wing": "has_weapon && soul_enemy(SOUL_ENEMY_STALFOS)"
-    "Forest Temple West Garden": "can_hit_triggers_distance"
-    "Forest Temple East Garden": "can_hit_triggers_distance"
+    "Forest Temple Initial Hallway With Skultullas Facing Player": "true"
+  events: #longshot_anywhere is for action shuffle and not being able to climb; redundant currently with can_hookshot
+    FOREST_MQ_ENTRANCE_ROOM_CHEST_SWITCH: "longshot_anywhere || climb_anywhere || (has_nuts || can_hit_triggers_distance || can_hookshot || has_hover_boots || can_use_din || has_bomb_bag || can_use_mask_stone || (has_weapon && (has(MAGIC_UPGRADE) || has(SHARED_MAGIC_UPGRADE))))" #Many ways to deal with the Skulltula in the way.
```
XenoWars: Don't need anywhere for longshot since you just longshot the vines.

XenoWars (changes requested):
There's comments regarding structuring of the rotating rooms, but I have a different idea about how to do them that I only thought of towards the end of the review, so i'll just leave them in so you can see my original thoughts too.

But I believe we can just structure them as West Rotating Room Hallway/Ledge/Pit and East Rotating Room Hallway/Ledge/Pit. Pit being the central part of the room after dropping down (but not in the holes). Would make things a lot neater and compact. The boss key chest and exits to Floormaster and Checkerboard would be moved to their respective pit regions with the necessary requirements, and then longshot/climbing up to them would go into the pit region.

packages/data/src/macros/macros_oot.yml
```diff
@@ -203,7 +203,9 @@
 "can_rescue_carpenter": "small_keys_hideout_all && (can_use_sword || ((can_boomerang || has_nuts) && can_use_sticks)) && soul_carpenters && soul_thief_fighter"
 "carpenters_rescued": "setting(gerudoFortress, open) || (event(CARPENTER_1) && (setting(gerudoFortress, single) || (event(CARPENTER_2) && event(CARPENTER_3) && event(CARPENTER_4))))"
 "get_past_skultullas": "has_nuts || can_use_sword_or_sticks || has_ranged_weapon || can_use_din || has_explosives || can_hammer || can_use_mask_blast || can_use_mask_stone"
+"get_past_skultullas_no_stone": "has_nuts || can_use_sword_or_sticks || has_ranged_weapon || can_use_din || has_explosives || can_hammer || can_use_mask_blast"
```
XenoWars, line 206: Oh yeah, and you can also roll to get behind them, at least in forest mq

XenoWars (changes requested):

packages/data/src/world/mq/forest_temple_mq.yml
```diff
@@ -14,137 +11,437 @@
...
+    "Forest Temple Gardens Well": "(has_tunic_zora && has_iron_boots) || event(FOREST_MQ_DRAIN_WELL)"
+    "Forest Temple West Garden Ledge Above Well": "true"
+    "Forest Temple West Garden Webbed Alcove": "event(FOREST_MQ_WEST_GARDEN_HIGH_WEB_DESTROY)" #Maybe another way to destroy web with glitches?
+    "Forest Temple East Garden Ground Area": "has_iron_boots || (has_tunic_zora && can_dive_big)" #Not part of Well region because diving with scale can't get checks. 
+  events:
+    FOREST_MQ_WEST_GARDEN_HIGH_WEB_DESTROY: "has_fire_arrows"
+"Forest Temple West Garden Webbed Alcove":
+  dungeon: Forest
+  exits:
+    "Forest Temple West Garden Ground Area": "trick(OOT_PASS_COLLISION) || event(FOREST_MQ_WEST_GARDEN_HIGH_WEB_DESTROY)" #Can be "true" if invisible web?
```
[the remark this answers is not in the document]
XenoWars: Ghun has provided the info that they are completely visible and solid.

packages/data/src/world/oot/dungeons_mq/forest_temple_mq.yml
```diff
@@ -62,23 +62,26 @@
...
   dungeon: Forest
   exits:
     "Forest Temple Main Lobby": "true"
-    "Forest Temple Lower Section Of Stalfos Battle Room": "is_child || (is_adult && can_play_time)" #What about spawning here with SoT block and going backwards through it?
+    "Forest Temple Lower Section Of Stalfos Battle Room": "is_child || can_play_time" #Will need more handling for the SoT block when door rando logic is added.
 "Forest Temple Lower Section Of Stalfos Battle Room":
   dungeon: Forest
   exits:
-    "Forest Temple Hallway Before Lower Section Of Stalfos Battle Room": "event(FOREST_MQ_UNDER_STALFOS_BATTLE_WOLFOS_CLEAR)"
-    #"Forest Temple Upper Section Of Stalfos Battle Room": "false" #Can't get up there when the Stalfos move the floor down.
```
Joshua8600: I prefer leaving this kind of connection written in, despite being commented out, in case some kind of OoB glitch is ever found to make this connection actually possible so it would be easy to just add the glitch in.

Issue #399, opened 2023-04-30
Bug: MM Mask Get Item Fanfare sometimes kills BGM


Revvenn:
This is a very minor bug that I've known about for awhile, just didn't note it down anywhere. But basically, if you get a mask from a chest whether it's in the overworld or in a dungeon it will sometimes completely kill the BGM so you will have no background music until you leave the area. It doesn't cause any gameplay issues but it's something MMR had for awhile early on as well.

I am sure this might get fixed or be easily fixable when Owl is done merging the audio tables. But I just thought I'd note this down for now.

Revvenn:
This is more likely to be tied to the audio heap not being increased in size. See https://github.com/OoTMM/OoTMM/issues/579

Revvenn:
I guess this should be stated now as: _any_ fanfare besides the minor item fanfare can kill BGM. If you get a small key, a boss key, heart container, heart piece, any major item, or any mask. It's because (I believe it is, anyway) the sequence that is playing when merged with the fanfare is too large so the audio for BGM breaks. This issue doesn't exist in MMR and is _supposedly_ not an issue anymore in ZOOTR. I am unsure what you can do to fix it when you already applied the above fix and applied the day 1 fix when you first enter Clock Town. But neither of the solo randos have this happen anymore either way.

Revvenn:
Alright, thanks to Lulu we have properly identified what the issue with this is:

> It's not an instrument issue, it's a sequence overtaking the sequence player from another sequence. OOTMM doesn't have anything in place to prevent this currently. So BGM songs that use the fanfare sequence player — like the aforementioned song — will stop playing as soon as a fanfare using the fanfare sequence player loads. And vice versa.

It's a quick and easy fix on the creators part (or users if they wanna bother with it), just changing a number in the .bankmeta file using a hex editor: https://discord.com/channels/1004394204992118935/1173446687214293053/1305684348497821749

Source: https://discord.com/channels/1004394204992118935/1173446687214293053/1327177949488742442

[...]

PR #849, opened 2025-04-24
Band-Aid Fix for Custom Music Bankmeta and Sequence Player Mismatches

crinuleiroz:
This is a band-aid that should partially fix issue #399 — it's a band-aid and only partially because it only affects the bank metadata for custom music, not vanilla music (I am unsure if OOTMM adds them into the pool when using custom music though).

Before the bank metadata array gets injected for a custom music file, the sequence player and song type get checked, and if they do not match then the sequence player in the array gets set to the matching sequence player.

> [!IMPORTANT]
> There's a special case for Kaepora Gaebora, because, despite being BGM, it *should* use the fanfare sequence player.
>
> The following songs intentionally mismatch the song type and sequence player:
> - Kaepora Gaebora's Theme (`NA_BGM_OWL`)
> - Keaton's Theme (`NA_BGM_KEATON_QUIZ`)
> - Kamaro's Theme (`NA_BGM_KAMARO_DANCE`)


Nax:
Thanks for the patch.
What about songs that use a vanilla bank? Am I correct in assuming it doesn't fix these?

crinuleiroz:
Yes, it's only custom songs, which is why it's just a band-aid fix. To fix vanilla songs you would need to get the metadata for the target song, then match the replacement song's metadata with it.

If I understand correctly that's what MMR does with this ASM patch and function:
- Patch in [music.asm](https://github.com/ZoeyZolotova/mm-rando/blob/3aceffa2d91ac76a9546004c3a7cb5142fbbf498/assembly/src/Music.asm#L46-L60) (there are two hooks)
- `Music_GetAudioLoadType` in [music.c](https://github.com/ZoeyZolotova/mm-rando/blob/3aceffa2d91ac76a9546004c3a7cb5142fbbf498/assembly/c/Music.c#L239-L247)

Nax:
Allright thanks, will try to implement the more durable solution eventually

PR #856, opened 2025-05-12
More preset updates

Revvenn:
-All Blitz presets have easier sun blocks enabled

-Allsanity and Hell presets have the newer shuffles added (Red Boulders, Red Ice, Icicles, Frog Rupees in OOT, and OOT Merchants)

-Updated Crosskeys to have the new shared items it was missing (Hammer, Spin Attack Upgrade), Fast Shadow Temple Boat, Easier Sun Blocks, added starting with MM Nuts and Sticks for now until figuring out if Keep Items on Cycle Reset should be enabled (and thus allowing for Shared Sticks and Nuts without losing them).

I will continue to keep presets updated as I am able.


Revvenn:
Oh I guess I did this slightly incorrect by not pulling the latest stuff first. Oops.

PR #876, opened 2025-08-25
Fixed Lab Dive

Ghunzor:
Changed the previous logic to can_dive_big to include bronze scale setting

### DIFF
--- a/packages/data/src/world/oot/overworld/lake_hylia.yml
+++ b/packages/data/src/world/oot/overworld/lake_hylia.yml
@@ -110,7 +110,7 @@
   exits:
     "Lake Hylia": "true"
   locations:
-    "Laboratory Dive": "soul_scientist && (has_scale_raw(2) || (trick(OOT_LAB_DIVE_NO_GOLD_SCALE) && has_iron_boots && can_hookshot && can_swim))"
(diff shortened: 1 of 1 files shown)


packages/data/src/world/oot/overworld/lake_hylia.yml
```diff
@@ -110,7 +110,7 @@
   exits:
     "Lake Hylia": "true"
   locations:
-    "Laboratory Dive": "soul_scientist && (has_scale_raw(2) || (trick(OOT_LAB_DIVE_NO_GOLD_SCALE) && has_iron_boots && can_hookshot && can_swim))"
+    "Laboratory Dive": "soul_scientist && (can_dive_big || (trick(OOT_LAB_DIVE_NO_GOLD_SCALE) && has_iron_boots && can_hookshot && can_swim))"
```
CelestialKitsune: `can_dive_big` expands to: "has gold scale or has iron boots" which is not what we want here.

```suggestion
    "Laboratory Dive": "soul_scientist && (cond(setting(bronzeScale), has_scale_raw(3), has_scale_raw(2)) || (trick(OOT_LAB_DIVE_NO_GOLD_SCALE) && has_iron_boots && can_hookshot && can_swim))"
```

PR #889, opened 2025-09-02
Jabu-Jabu MQ Bronze Scale Logic

Joshua8600:


packages/data/src/world/oot/dungeons_mq/jabu_jabu_mq.yml
```diff
@@ -31,24 +31,30 @@
   dungeon: JJ
   exits:
     "Jabu-Jabu Main Elevator Room Upper": "true"
-    "Jabu-Jabu Main Elevator Room Lower Switch Alcove": "is_child || (is_adult && can_dive_small)" #Adult cannot dive to the alcove without a scale.
-    "Jabu Jabu Basement Loop to Main Elevator": "event(JABU_MQ_WATER_SPOUTS)"
+    "Jabu-Jabu Main Elevator Room Lower Switch Alcove Ledge": "can_swim && (is_child || (is_adult && can_dive_small))" #Adult cannot dive to the alcove without a scale. However, can hookshot through Jabu's flesh to the compass chest without bronze scale (only irons, and zora tunic?). Can be a trick.
```
XenoWars: Hookshotting through Jabu's flesh has already been discussed and would be classified as glitch. It is similar to longshotting through King Zora or the valley boulders.

packages/data/src/world/oot/dungeons_mq/jabu_jabu_mq.yml
```diff
@@ -31,24 +31,30 @@
...
     "MQ Jabu-Jabu Second Room B1 Chest": "true"
     "MQ Jabu-Jabu Wonder Item Elevator Cow": "can_use_slingshot"
-"Jabu-Jabu Main Elevator Room Lower Switch Alcove":
+    "MQ Jabu-Jabu Rupee Bottom": "has_iron_boots" #Any requirement with swimming is better suited for the other region where these checks are.
+    "MQ Jabu-Jabu Rupee Middle": "has_iron_boots" #Any requirement with swimming is better suited for the other region where these checks are.
+    "MQ Jabu-Jabu Rupee Top": "has_iron_boots" #Any requirement with swimming is better suited for the other region where these checks are.
+"Jabu-Jabu Main Elevator Room Lower Switch Alcove Ledge":
   dungeon: JJ
   exits:
     "Jabu-Jabu Main Elevator Room Lower": "is_child || (is_adult && can_dive_small)" #Adult cannot dive to the alcove without a scale.
```
XenoWars: It doesn't matter, but this would also need can_swim

packages/data/src/world/oot/dungeons_mq/jabu_jabu_mq.yml
```diff
@@ -78,29 +84,39 @@
...
     "MQ Jabu-Jabu Grass Main Room Bottom 1": "can_cut_grass"
     "MQ Jabu-Jabu Grass Main Room Bottom 2": "can_cut_grass"
     "MQ Jabu-Jabu Grass Main Room Bottom 3": "can_cut_grass"
-"Jabu Jabu Basement Loop to Main Elevator":
+"Jabu Jabu Basement Loop to Main Elevator West Side":
   dungeon: JJ
   exits:
-    "Jabu-Jabu Ruto Section Bottom Floor": "true"
     "Jabu-Jabu Main Elevator Room Lower": "event(JABU_MQ_WATER_SPOUTS)" #We are going to need a buffer region here for behind the water spout. You can use Hover Boots with Bunny Hood or backwalking to get through the spout, or climb anywhere. There is not another one blocking the door on this side. It should also probably be a trick, even with climb anywhere.
+    "Jabu Jabu Basement Loop to Main Elevator East Side": "can_swim || can_hookshot || climb_anywhere"
```
XenoWars: Hover Boots work as well.

packages/data/src/world/oot/dungeons_mq/jabu_jabu_mq.yml
```diff
@@ -225,15 +241,25 @@
     "Jabu-Jabu Back Forked Paths": "true"
   locations:
     "MQ Jabu-Jabu GS Back": "gs && ((climb_anywhere && (can_hammer || can_use_mask_blast)) || (has_explosives && can_collect_distance))"
-"Jabu-Jabu Basement Side Room":
+"Jabu-Jabu Basement Side Room Entrance Side":
+  dungeon: JJ
+  exits:
+    "Jabu-Jabu Ruto Section Bottom Floor": "can_swim && event(JABU_MQ_BASEMENT_SIDE_BUTTON)"
```
XenoWars: This should just be the event.

packages/data/src/world/oot/dungeons_mq/jabu_jabu_mq.yml
Joshua8600: Added

packages/data/src/world/oot/dungeons_mq/jabu_jabu_mq.yml
Joshua8600: Comment edited to say glitch

packages/data/src/world/oot/dungeons_mq/jabu_jabu_mq.yml
Joshua8600: Added

packages/data/src/world/oot/dungeons_mq/jabu_jabu_mq.yml
Joshua8600: Not entirely. Because the event can just be iron boots (can_dive_small), but you need to swim to actually exit. And it's written like this for action shuffle, so it needs to be separated out like this

XenoWars: You can just unequip the boots to exit. Would be silly to require swimming just to leave the room when you can just press the switch with irons then unequip to void back to land.

Joshua8600: I don't know if the button persists with a void. It's not a permanent flag.

Joshua8600: I just checked. Button does not persist through a bronze scale void. So it needs can_swim

PR #904, opened 2025-09-29
Updated DC Logic

Ghunzor:
Added 
- Blast Mask logic to mud walls
- Added Region for Near Boss GS exit
- Added methods to kill GS logic for future door rando

### DIFF
--- a/packages/data/src/world/oot/dungeons/dodongo_cavern.yml
+++ b/packages/data/src/world/oot/dungeons/dodongo_cavern.yml
@@ -2,13 +2,13 @@
   dungeon: DC
   exits:
     "Death Mountain": "true"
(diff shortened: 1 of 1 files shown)


packages/data/src/world/oot/dungeons/dodongo_cavern.yml
```diff
@@ -95,7 +95,7 @@
     "Dodongo Cavern Stairs": "true"
     "Dodongo Cavern Bomb Bag Room 1": "true"
   locations:
-    "Dodongo Cavern GS Stairs Vines": "gs"
+    "Dodongo Cavern GS Stairs Vines": "gs && (can_collect_distance || has_ranged_weapon_child || has_ranged_weapon_adult || (is_adult && can_damage_skull) || has_explosives)" ## Check if biggoron and child can kill
```
XenoWars: This can be reverted. It is trivial because of the pots.

PR #919, opened 2025-10-23
Water MQ Blue Switch Path Fixes

Joshua8600:


packages/data/src/world/oot/dungeons_mq/water_temple_mq.yml
```diff
@@ -93,9 +93,11 @@
 "Water Temple Storage Room":
   dungeon: Water
   exits:
-    "Water Temple Main Room Middle Floor": "(event(WATER_LEVEL_MID) || event(WATER_LEVEL_LOW)) || (event(WATER_LEVEL_HIGH) && has_iron_boots && has_tunic_zora)"
+    "Water Temple Main Room Middle Floor": "((event(WATER_LEVEL_MID) || event(WATER_LEVEL_LOW)) && (is_adult || (can_swim || has_hover_boots || hookshot_anywhere)))  || (event(WATER_LEVEL_HIGH) && has_iron_boots && has_tunic_zora)"
```
XenoWars: For some reason, on every set of requirements starting with is_adult, every other option is included in an unneccesary separate set of parentheses.

Also, here specifically the logic is incorrect. It's going to be very difficult to swim through thin air, plus ice arrows not here. Thankfully nothing will ever break from it since the connection is useless and redundant but if it's desired to keep it, it should still be fixed.

XenoWars (changes requested):

packages/data/src/world/oot/dungeons_mq/water_temple_mq.yml
Joshua8600: Should be fixed now

XenoWars (approved):

PR #924, opened 2025-10-28
Add shared bottles.

ZoeyZolotova:
- Normal item pool has 6 bottles (Red Potion, Gold Dust, Chateau, Milk, Ruto Letter, Empty)
- Plentiful item pool has 7 bottles (Red Potion, Gold Dust, Gold Dust, Chateau, Milk, Ruto Letter, Ruto Letter) and the last one you find will do nothing.
- In OoT, the 2 extra bottles go on the adult trade and child trade slots. Child and Adult can both equip bottles from those slots regardless of age, but won't be able to cycle the slot if the item is equipped. If the bottle is equipped as adult while you're child or vice versa and you cycle the slot, the bottle is unequipped from the age where it's not appropriate.
- All bottle contents in the main 6 slots are shared between games.
- Add custom Blue Fire actor for MM.
- Add custom Water/Hot Water actor for OoT.
- In OoT, you can drink Chateau Romani, give a Mushroom to the old hag in Kakariko, and water magic bean plants with spring water or hot spring water to summon fairies. You can also sell mushrooms, chateau, spring water, hot spring water and zora eggs to the beggar for the same price as the curiosity shop takes them. Spring Water and Hot Spring Water destroy mudwalls just like Blue Fire does.
- In MM, you can sell blue fire to the curiosity shop for 50 rupees, read Ruto's Letter, give Ruto's Letter to the toilet hand, and use Blue Fire for no purpose. Blue Fire activates sun blocks just like spring water does.
- Both King Zora and the Toilet Hand will return Ruto's Letter to you if you haven't yet given it to the other. This allows Ruto's Letter to remain non-renewable. Each has altered text in those circumstances.
- Reading Ruto's Letter when shared bottles are enabled will add "PPS: Don't flush this letter!" to the end of it.
- Works with the Random Bottle Contents setting. Ruto's Letter and Gold Dust are kept, the rest are randomized.


packages/core/src/common/item/item_add.c
```diff
@@ -2185,6 +2232,37 @@ static const SharedItem kSimpleSharedItems[] = {
     { CFG_SHARED_NUTS_STICKS, GI_OOT_NUT_UPGRADE2, GI_MM_NUT_UPGRADE2 },
     { CFG_SHARED_STONE_OF_AGONY, GI_OOT_STONE_OF_AGONY, GI_MM_STONE_OF_AGONY },
     { CFG_SHARED_SPIN_UPGRADE, GI_OOT_SPIN_UPGRADE, GI_MM_SPIN_UPGRADE },
+    { CFG_SHARED_LENS, GI_OOT_POTION_RED, GI_MM_POTION_RED },
+    { CFG_SHARED_LENS, GI_OOT_POTION_GREEN, GI_MM_POTION_GREEN },
+    { CFG_SHARED_LENS, GI_OOT_POTION_BLUE, GI_MM_POTION_BLUE },
+    { CFG_SHARED_LENS, GI_OOT_FAIRY, GI_MM_FAIRY },
+    { CFG_SHARED_LENS, GI_OOT_BOTTLE_MILK, GI_MM_BOTTLE_MILK },
```
Nax: Wrong cfg, should be SHARED_BOTTLES here

ZoeyZolotova: Oh gosh, thank you.

PR #936, opened 2025-11-03
Updated Kak Well Exit

Ghunzor:
Added longshot anywhere with iron boots

### DIFF
--- a/packages/data/src/world/oot/overworld/kakariko.yml
+++ b/packages/data/src/world/oot/overworld/kakariko.yml
@@ -114,7 +114,7 @@
 "Kakariko Well":
   region: KAKARIKO
   exits:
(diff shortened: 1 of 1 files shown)


packages/data/src/world/oot/overworld/kakariko.yml
```diff
@@ -114,7 +114,7 @@
 "Kakariko Well":
   region: KAKARIKO
   exits:
-    "Kakariko": "can_swim || is_adult"
+    "Kakariko": "can_swim || is_adult || (is_child && longshot_anywhere)"
```
XenoWars, line 117: It does not need to specify child here. It is implied by not being adult.

Nax: Doesn't really harm either

PR #937, opened 2025-11-09
Fixed raise pillar event (1)

Ghunzor:
Updated the event to correct logic.
Removed || and added &&

### DIFF
--- a/packages/data/src/world/mm/dungeons/snowhead_temple.yml
+++ b/packages/data/src/world/mm/dungeons/snowhead_temple.yml
@@ -126,7 +126,7 @@
     "Snowhead Temple Pillars Room Lower": "true"
     "WARP_SONGS": "true"
   events:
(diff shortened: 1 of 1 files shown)


packages/data/src/world/mm/dungeons/snowhead_temple.yml
```diff
@@ -126,7 +126,7 @@
     "Snowhead Temple Pillars Room Lower": "true"
     "WARP_SONGS": "true"
   events:
-    SNOWHEAD_RAISE_PILLAR: "(has_mask_goron || can_hammer) && (can_use_fire_short_range || (event(SHT_STICK_RUN) && trick(MM_SHT_STICKS_RUN)) || trick_sht_hot_water_er)"
+    SNOWHEAD_RAISE_PILLAR: "(has_mask_goron || can_hammer) && (can_use_fire_short_range || ((event(SHT_STICK_RUN) && trick(MM_SHT_STICKS_RUN) && trick_sht_hot_water_er)))"
```
XenoWars, line 129: One set of () too many around event and er, but would not break

PR #951, opened 2026-02-28
Add every ocarina song to both games. Add Song Events Shuffle (MM).

ZoeyZolotova:
- Add Song of Soaring (OoT) and Ageless Song of Soaring to replace Cross-Games MM Song of Soaring.
- Add Sonata of Awakening (OoT), Goron Lullaby (OoT), New Wave Bossa Nova (OoT), Oath to Order (OoT), which do nothing unless Song Events Shuffle is on.
- Add Song of Healing (OoT), which heals signs and summons fairies at gossip stones.
- Add Zelda's Lullaby (MM), which heals signs and summons fairies at gossip stones.
- Add Saria's Song (MM), which makes some redeads dance and summons fairies at gossip stones.
- Add Minuet of Forest (MM), Bolero of Fire (MM), Serenade of Water (MM), Requiem of Spirit (MM), Nocturne of Shadow (MM) and Prelude of Light (MM) to replace Cross-Games OoT Warp Songs.
- Add Shared settings variants of each song above.
- Rename Sun's Song in MM to Sun's Song (MM) to be consistent.
- Fix frog game softlocking when playing a custom song.


packages/core/src/settings/data.ts
```diff
@@ -2054,6 +2046,126 @@ export const SETTINGS = [{
   description: "Add the Elegy of Emptiness in Ocarina of Time. If Songs are on Song Locations, you must share or start with at least one song",
   default: false,
   cond: hasOoT,
+}, {
+  key: 'songSoaringOot',
+  name: "Song of Soaring (OoT)",
+  category: 'items.extensions',
+  type: 'boolean',
+  description: "Add the Song of Soaring in Ocarina of Time. Usable only as child. If Songs are on Song Locations, you must share or start with at least one song",
```
Onemario1234, line 2054: "you must share or start with at least one song" is a lot less factual when taking into account double the songs of the original games, assuming not shared. Perhaps these should be clarified and expanded on.

ZoeyZolotova: Any suggestions?

Onemario1234: Best I got is something along the lines of

"Any extra songs added without being shared will require Songs shuffled Anywhere, as too many will exist for Song Locations"

That can probably be shortened in some way, maybe by removing the Song Locations mention, but that's what I was able to think of on the spot.

Revvenn:
Only comment I have is ensuring Song of Soaring, Double Time, and the newly added MM songs don't softlock when playing to the frogs in Zora's River (as currently, they do, at least Soaring/Double Time do currently in current dev). Unfortunately this has to be addressed because someone _did_ try this and they got softlocked.

ZoeyZolotova:
> Only comment I have is ensuring Song of Soaring, Double Time, and the newly added MM songs don't softlock when playing to the frogs in Zora's River (as currently, they do, at least Soaring/Double Time do currently in current dev). Unfortunately this has to be addressed because someone _did_ try this and they got softlocked.

Fixed it, thanks for the reminder.

Revvenn:
Oh actually there *was* one more thing: the newly added songs need to be added to the file select screen like Elegy in OOT currently is handled. As in, the song sprites on the file select screen to show whether you have the songs for that game and such.

Remembered this earlier today lol.

packages/core/src/settings/data.ts
```diff
@@ -1720,17 +1708,29 @@ export const SETTINGS = [{
...
-  description: 'Alters the behavior of the Goron Lullaby',
+  description: 'Alters the behavior of the MM Goron Lullaby',
   values: [
     { value: 'single', name: 'Full Lullaby Only', description: 'Only the Goron Lullaby can be found, and when playing with Songs on Song Locations, Baby Goron in MM is no longer a Song Location' },
     { value: 'progressive', name: 'Progressive', description: 'The Lullaby Intro will be received first before getting the full song' },
   ],
   default: 'progressive',
   cond: hasMM,
+}, {
+  key: 'progressiveGoronLullabyOot',
```
Nax, line 1723: I don't think it makes much sense to have two different settings for this, a shared one ought to be enough?

Onemario1234: I was gonna try and offer some thoughts on why it could be necessary, but I don't have much aside from the fact that I just think it's kinda neat that they're separate. I won't argue against giving them a combined progressive setting that works for them both, kinda like the Blast Mask Cooldown setting, though I do like the ability to change them separately.

packages/generator/src/mm/actors/Dm/Dm_Stk.c
```diff
@@ -92,7 +92,7 @@ static void hintOathToOrderNotes(PlayState* play)
 
 static void DmStk_Start(Actor* this, PlayState* play)
 {
-    if (Misc_IsMoonAccessible() && !Config_Flag(CFG_MM_OPEN_MOON))
+    if (Misc_IsMoonAccessible() && !Config_Flag(CFG_MM_OPEN_MOON) && !Config_Flag(CFG_MM_SONG_EVENTS_SHUFFLE))
```
Nax, line 95: IMO the hint should just hint whatever song it is that lock the moon instead

Nax: Not that it's super blocking for now

Nax (approved):

Nax:
(Merging through a merge branch since you merged master instead of rebasing, to rebuild custom.h)

Nax:
Actually custom.h is up to date

PR #962, opened 2026-05-04
Animation importer and Kamaro Mask

Chronolegionnaire:
Initial implementation of animation importing pipeline. 

Link/Player animations can be ported freely between games using the linkanimations.yml file.

Frame loading patched to allow loading animation data outside of link_animetion to get around limitation of not being able to expand segment 7 vrom space.

(I couldn't get headers to properly populate in each game so a work around is used to populate header information on demand)

This should hopefully make other item ports much more straightforward in the future. 

Kamaro mask ported to oot to test foreign imported animations. Kamaro mask behavior fully (for the most part) recreated in oot. On the MM side I removed the hammer animation patches and new ported hammer animations are used directly and tested successfully with the new animations.


Chronolegionnaire:
One thing I wasn't really sure of when I was building this was actual cart space budget there is to work with for each game. The animation data isn't huge but it could potentially add up, if for example all the transformation mask animations were ported to oot. I set the file address for the new animation stores really high hoping it would just be placed at the end of each rom, which was successful but also not 100% confident those storage spaces are totally safe. (I'm still learning the nuances of how addressing and space stuff actually works on n64) They work fine in context of the 1 animation for oot and 6 for mm in the current pr, but not sure if there will be side effects if more get ported.

packages/generator/lib/combo/names.ts
ZoeyZolotova, line 1: should also change MM_MASK_KAMARO to be "Kamaro Mask (MM)"

packages/generator/lib/combo/presets.ts
```diff
@@ -116,7 +118,8 @@ const BLITZ_BASE: PartialDeep<Settings> = {
     MM_OCARINA: 1,
     OOT_OCARINA: 1,
     MM_SWORD: 1,
-    MM_SONG_SOARING: 1
+    MM_SONG_SOARING: 1,
+    SHARED_MASK_KAMARO: 1,
```
ZoeyZolotova: Was this just for testing?

Chronolegionnaire: Yep didn't mean to leave that in

Chronolegionnaire:
Okay I actually think I got it to dma foreign data so that completely eliminates the worry about taking up more rom space (for some reason I completely missed the already existing foreign dma function), new version just makes a custom header that references the foreign data instead, cleaning it up but should have another push for this pr either later tonight or tomorrow.

PR #964, opened 2026-05-06
Cross animation

Chronolegionnaire:
Okay 3rd times the charm, cross animation stuff only


Chronolegionnaire:
I don't think a I can do a pr based on this pr without also including all the changes from this one into master unless this one gets converted to a branch so the kamaro mask branch will have to wait until this pr either gets pushed to a new branch and/or approved for merge assuming its acceptable.

Chronolegionnaire:
The last thing I think this needs before a solid review is finding a way to make it actually resolve the custom header, all the things I try keep failing to resolve animation header so I ended up doing a static list of animation symbols that it checks against. Which for a small number of animations is probably fine, but if tons get imported and the list grows that will also add up and may cause performance issues.

Chronolegionnaire:
Okay ended up ditching gameplay_keep expansion completely and just using a custom header but now it has parity with how vanilla animations are processed

data/defs/link-animations.yml
```diff
@@ -0,0 +1,35 @@
+animations:
```
Nax, line 1: This file should be converted to json in the pre-processing step, so that we don't have a yaml dependency at runtime (remember, this will be bundled for web later, no filesystem there)

packages/generator/lib/combo/custom/custom-animation-builder.ts
```diff
@@ -0,0 +1,426 @@
...
+    lines.push('# define COMBO_IMPORTED_LINK_ANIMATION_LIST(_)');
+    lines.push('#endif');
+    lines.push('');
+    lines.push('#endif');
+    lines.push('');
+
+    await atomicWriteFile(path.join(BUILD_INCLUDE_DIR, 'imported_animations.h'), lines.join('\n'));
+}
+
+export async function patchAnimationPorts(roms: DecompressedRoms) {
```
Nax, line 315: Ditto, this cannot use the filesystem because this part runs on the web.

packages/generator/src/oot/actors/Animation/Animation.c
```diff
@@ -0,0 +1,350 @@
...
+typedef void (*LinkAnimationEndLoopFn)(SkelAnime* skelAnime);
+
+typedef s32 (*LinkAnimationOnFrameFn)(
+    SkelAnime* skelAnime,
+    f32 frame
+);
+
+/* Player wrappers */
+
+void Player_SetupActionPreserveItemAction(
```
Nax: These wrappers don't seem very useful, could just be added to `link_oot`/`link_mm`

Nax (commented):
Looks okay, some minor adjustemnts to make, I'll see what I can do

PR #965, opened 2026-05-09
Keep music name visible while paused

Sharpoint2:
When you pause the game, the name of the song playing (assuming you have music randomized) will display again when the game is paused.

### DIFF
--- a/packages/generator/src/common/audio.c
+++ b/packages/generator/src/common/audio.c
@@ -378,6 +378,16 @@ void Audio_DisplayMusicName(PlayState* play)
     if (!sDisplayMusicNames)
         return;
 
+#if defined(GAME_OOT)
+    /* Don't show music name on the Song of Soaring map screen */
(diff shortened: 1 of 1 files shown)


CelestialKitsune:
Regarding the use of `play->pauseCtx.state != 0` - I fear it might also show when a death happens or when you soar, given `0` is `PAUSE_STATE_OFF` in both decomps, meaning that any other pause states are possible candidates to pass the condition

Sharpoint2:
It doesn't appear when you die in either game
But it does appear when soaring in MM. (I forgot to enable cross game soaring this seed so I haven't had a chance to check OOT yet)

Nax:
IIRC we explicitely disabled it in the pause menu because it caused weirdness & crashes on actual hardware (because of 1cycle/2cycles shenanigans).
Have this been tested on at least Ares?

Nax:
Let's merge and see if hardware players report issues

PR #967, opened 2026-05-11
Boomerang port

Chronolegionnaire:
MM Boomerang port v1


Chronolegionnaire:
Current issues:

~~The thrown object doesn't look quite right to me, it could still us some tweaks to make it match the oot look a little more~~

Fixed

~~Equip animation currently uses zora fin equip/put away animation, would take some extensive patching and/or custom equip animation code to fix this so i left it as is for now~~
Fixed hand models, equip animation is still a right handed item instead of a left handed item like bottle

Camera currently uses CAM_MODE_ZORAFIN. CAM_MODE_BOOMERANG does exist and works in overworld but breaks in dungeons because (i suspect) its not in the dungeon camera tables. Currently the patch to enable it is commented out.

~~Icon in inventory always has the pictobox picture count because they share a slot. Needs some patching on the inventory ammo count display to make that go away.~~
Fixed

~~Shared item doesn't grant the mm boomerang~~
Fixed

Chronolegionnaire:
The remaining two issues are very minor cosmetic issue that will take some extensive patching to fix:
Wrong equip/put away animation (right hand instead of left hand)
uses CAM_MODE_ZORAFIN instead of CAM_MODE_BOOMERANG

PR #969, opened 2026-05-15
Oot gfs

Chronolegionnaire:
This implementation is kind of hacky, and wouldn't work for any other item but I wanted to create the PR to see what ya'll think. Its basically spoofing the new GFS item as the biggoron sword which works because they are functionally the same item. Outside of powder keg (which there's not really a reason to port without goron mask anyway) there's no other non mask items to port anyway so there is not much to gain out of building a custom IA  pipeline like the MM side has (unless some brand new custom items were ever added). This works 1 to 1 to MM GFS but has one issue:

1: Because they share the same IA pointer, if you equip both Biggoron Sword and have GFS on a c action and have one or the other out. Using either will swing the current weapon that's out.

Still needs some logic additions but I have added the shared item checks to mm so it doesn't break generation.


Chronolegionnaire:
This would benefit greatly from Zoey's bomb bag pr so it might be best to wait to merge this until after that is reviewed and merged, and I can rebase it and rebuild gfs as its actual own item rather than just spoofing it as biggoron. (which would fix the aforementioned bug it has right now)

Nax:
Okay let's wait then

Chronolegionnaire:
> Okay let's wait then

With the powder keg pr I was able to update this to move GFS into a standalone item, this should be ready for a proper review and merge now

PR #976, opened 2026-05-25
Boomerang fixes

Chronolegionnaire:
Fix boomerang not being able to interact with bio baba stems because of lack of func_800B90AC collision check

Fix item00 deleting itself when boomerang carries it out of bounds by setting gravity to 0


Chronolegionnaire:
Not sure why it keeps pushing 5 commits, I'm just going to full reset my branch to hopefully fix this

ZoeyZolotova (approved):

PR #982, opened 2026-05-29
Song event plando

Chronolegionnaire:
Song event shuffle plando

Features:

Allows to plando any song event as single plando, or group plando.

Random assignment: Mainly exists for groups so that you can make an entire group of events all have the same random song. Otherwise assigning it to a single song is the same as not assigning anything at all to the event in which case it will shuffle as normal.

Tab is hidden if no song event shuffle settings are on.

Songs and event hide dynamically based on what items are available so you won't be able to assign an item that doesn't exist in one game to an event in that game.

Not super happy with the tool tip location, but couldn't really find a better place for it, and I do think the clarification is needed.

image

### DIFF
--- a/packages/core/src/settings/util.ts
+++ b/packages/core/src/settings/util.ts
@@ -8,7 +8,7 @@ import { DEFAULT_TRICKS, TRICKS } from './tricks';
 import { DEFAULT_SPECIAL_COND, DEFAULT_SPECIAL_CONDS, SPECIAL_CONDS, SPECIAL_CONDS_FIELDS } from './special-conds';
 import { patchArray } from './patch';
 import { SETTINGS_DEFAULT_HINTS } from './hints';
-import { SongEventSongs, SONG_EVENT_LOCATIONS_OOT, SONG_EVENT_LOCATIONS_MM } from '../song-events';
+import { SongEventSongs, SONG_EVENT_LOCATIONS_OOT, SONG_EVENT_LOCATIONS_MM, type PlandoSongEvent } from '../song-events';
 
 export const DEFAULT_SETTINGS: Settings = { ...SETTINGS.map(s => {
   if (s.type === 'set') {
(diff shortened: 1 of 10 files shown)


packages/core/src/settings/util.ts
```diff
@@ -174,6 +181,65 @@ function sortCopyObject<T extends {[k: string]: any}>(obj: T): T {
   return result as T;
 }
 
+function sortCopySongEvents<T extends readonly string[]>(
+    obj: unknown,
```
CelestialKitsune: Really not a fan of `unknown` - it should be avoided if possible.

packages/core/src/settings/util.ts
```diff
@@ -174,6 +181,65 @@ function sortCopyObject<T extends {[k: string]: any}>(obj: T): T {
...
+    const song = (data as { song?: unknown }).song;
+    const group = (data as { group?: unknown }).group;
+
+    const resultData: {
+      song: SongEventSongs | 'random';
+      group?: string;
+    } = {} as {
+      song: SongEventSongs | 'random';
+      group?: string;
+    };
```
CelestialKitsune: This should be refactored to use the type defined in `packages/core/src/song-events.ts`

CelestialKitsune (changes requested):
Style wise, ensure you're using LF instead of CRLF, and have new lines at the end of your files.

Most of the logic seems fine to me, there are parts that go over my head, so I'd prefer having an additional review

packages/core/src/settings/util.ts
Chronolegionnaire: Fixed in the latest commit!

packages/core/src/settings/util.ts
Chronolegionnaire: Fixed in the latest commit!

Chronolegionnaire:
> Style wise, ensure you're using LF instead of CRLF, and have new lines at the end of your files.
> 
> Most of the logic seems fine to me, there are parts that go over my head, so I'd prefer having an additional review

Regarding this, I am using LF, should I switch to CRLF?

CelestialKitsune (approved):

CelestialKitsune:
> Regarding this, I am using LF, should I switch to CRLF?

No no - I just saw a full file being diffed when only one or two line were added, making it seem like the line returns were changed from `\n` to `\r\n`

CelestialKitsune:
That said, squashing and merging, so an eventual revert will be easier.

Chronolegionnaire:
> That said, squashing and merging, so an eventual revert will be easier.

My git has been really wonky, not sure why it keeps doing that

PR #985, opened 2026-05-30
Potentially fix song event plando and add spoiler log entry for song events

Chronolegionnaire:
The song events were not indexing correctly with the actual events available in code because of the commented out entries (that I had commented out cause they were otherwise unused for actual event stuff as of right now). At least I think, I am honestly not 100% sure this is the cause yet, but after uncommenting the song events in the plando code I was able to generate a dozen or so seeds in a row and verified every event was correctly set to what it was plando'd for. With the unused events commented out, it would plando the songs incorrectly like 1 in 5 or so generations. Which is what confuses me, I would assume if indexes were misaligned it would always generate wrong. But this does seem to have fixed it? (Its possible I just got lucky and didn't actually run into the bug after uncommenting so it needs more testing)

Also added song event spoiler output so I could actually inspect what the the code thought should go where to have something to compare to in game for random groups.


CelestialKitsune:
Does the `plando` section also prints out what was set? If not, it should, on top of having a full section for song events

Chronolegionnaire:
> Does the `plando` section also prints out what was set? If not, it should, on top of having a full section for song events

Done and done!

PR #986, opened 2026-06-02
QoL Setting: Bow and Slingshot break Beehives in OoT

keraion:
This adds a QoL ability to break beehives with the slingshot or bow in OoT to mirror MM's behavior. It's added into logic via the two `break_hive_low` and `break_hive_high` macros which slingshot and bow can reach.

Let me know if there's anything I need to change or I have missed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


CelestialKitsune:
Force pushed to fix conflicts and update changelog. Waiting on pipeline to merge, but should be good to go

PR #988, opened 2026-06-02
Add setting for MM-style air movement in OoT.

ZoeyZolotova:


Nax:
Might be worth looking at potential logic changes when this is on (maybe you can bomb jump to water boss? stuff like that)

PR #998, opened 2026-06-07
Fix custom music not using formmask files when available.

ZoeyZolotova:
The JSON5 package is to parse json with trailing commas in arrays.


Nax:
JSON5 is pretty horrible, IMO either we accept YAML, or we stick to JSON and reject these malformed files

ZoeyZolotova:
> JSON5 is pretty horrible, IMO either we accept YAML, or we stick to JSON and reject these malformed files

Trailing commas are accepted by MMR's json serializer and a lot of custom songs with formmasks have it, including many of my own. It's not desirable to have everyone replace their affected custom songs with "fixed" versions.

If you like, I can remove JSON5 and instead manually fix the raw string with a find and replace before parsing it to json.

PR #1001, opened 2026-06-11
ootmaskMM

Chronolegionnaire:
Add the three missing oot masks to MM:

Gerudo Mask: Can be used as a disguise for pirates fortress

Skull Mask: Can be used as a disguise for Deku Palace (Maybe have custom dialogue for this in the future?)

Spooky Mask: Acts exactly like gibdo mask but only at night.


packages/generator/include/combo/mm/save.h
```diff
@@ -616,4 +619,14 @@ MmCustomSave;
...
 
+#define GET_WEEKEVENTREG(index) \
+(*(u8*)((u8*)&gSaveContext.save + 0xEF8 + (index)))
+
+#define CHECK_WEEKEVENTREG(flag) \
+(GET_WEEKEVENTREG((flag) >> 8) & ((flag) & 0xFF))
+
+#define WEEKEVENTREG_12_40 ((12 << 8) | 0x40)
+#define WEEKEVENTREG_17_04 ((17 << 8) | 0x04)
+#define WEEKEVENTREG_23_20 ((23 << 8) | 0x20)
```
CelestialKitsune: We have a whole file setup for events: `packages/generator/include/combo/common/events.h`, and for checking weekeventregs

packages/generator/src/mm/ovl_patch/kaleido_mask.S
```diff
@@ -0,0 +1,9 @@
+#include <combo.h>
+
+PATCH_START 0x8082b68a
+  .short 0
+PATCH_END
+
+PATCH_START 0x8082b6aa
+  .short 0
```
CelestialKitsune, line 10: Please add a comment why they're needed so we wreck our brains to know why this was added later on (admittedly all asm patches like that should explain, but that's a later problem)

CelestialKitsune (changes requested):

packages/generator/src/mm/ovl_patch/kaleido_mask.S
Chronolegionnaire: Comment added

packages/generator/include/combo/mm/save.h
Chronolegionnaire: Moved my event registrations to events.h and fixed call sites

CelestialKitsune (approved):

CelestialKitsune:
I don't see much else that could need to be changed, but it's a big PR, which makes me wary of things potentially breaking, despite the testing.
I'll merge still, but let's try to have smaller PRs from now on, if possible!

PR #1008, opened 2026-06-17
Changed deku seed drop restriction to an array for easy expansion

Chronolegionnaire:
Change the deku seed location check to an array for easy expansion

### DIFF
--- a/packages/generator/src/mm/actors/En/En_Item00.c
+++ b/packages/generator/src/mm/actors/En/En_Item00.c
@@ -257,23 +257,17 @@ void EnItem00_DrawShield(PlayState* play)
     Draw_Gi(play, NULL, gi, DRAW_RAW);
 }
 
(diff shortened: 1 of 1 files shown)


packages/generator/src/mm/actors/En/En_Item00.c
```diff
@@ -257,16 +257,35 @@ void EnItem00_DrawShield(PlayState* play)
...
+{
+    s32 i;
+
+    for (i = 0; i < ARRAY_COUNT(DekuSeedBlockedScenes); i++) {
+        if (DekuSeedBlockedScenes[i] == sceneId) {
+            return true;
+        }
+    }
+
+    return false;
```
CelestialKitsune: Preferred to have a switch and we add cases as needed, rather than having a loop and a static array

Chronolegionnaire: Updated

PR #1013, opened 2026-07-03
OOT Mask MM Fixes

Chronolegionnaire:
Removed all the moon child stuff because it was unsafe. ~~With this, handing in stone or gibdo mask while owning any of the three new masks will remove the new masks as well but they come back after a cycle reset which is preferred behavior to the inventory corrupting. I could use help figuring out a way to solve this as nothing I try seems to work.~~

All fixed!


packages/generator/src/mm/ovl_patch/kaleido_scope.c
```diff
@@ -630,23 +710,168 @@ void KaleidoScope_DrawIconCustom(GraphicsContext* gfxCtx, u8 item, u16 width, u1
     }
 }
 
+#define MOON_MASK_BIT(i, f) ((u16)(((i) << 8) | (f)))
+#define MOON_MASK_BYTES     ((u8*)0x801f3f3a)
```
CelestialKitsune: I am fairly sure this is in the save context, so rather than using a hardcoded address, I'd rather have the field be added in there.

CelestialKitsune (changes requested):
Overall seems fine to me. Thanks for looking into this

packages/generator/src/mm/ovl_patch/kaleido_scope.c
Chronolegionnaire: Updated

PR #1024, opened 2026-08-15
Fix very minor soul name typos/inconsistencies

eedefeed:
Three changes.

# toto
missing space

# larvaes -> larvae
"larvae" is already plural

# bean salesman
changed for consistency with OoT and shared souls.


eedefeed:
These changes folded into: https://github.com/OoTMM/OoTMM/pull/1021

PR #1025, opened 2026-08-15
Skull Mask treated like stone mask for bats, guays, and keese

Chronolegionnaire:
Hide player from bats, guays, and keese when wearing skull mask like in OoT.


Nax:
Has it been tested? Looks rather safe

Chronolegionnaire:
> Has it been tested? Looks rather safe

Tested with all three and seems to work like a charm. Did not inspect logic though, not sure if theres any logic implications for this.

PR #1028, opened 2026-08-16
mm adult mask

Chronolegionnaire:
MM Adult mask decoupled from age requirements


packages/generator/lib/combo/custom/index.ts
```diff
@@ -269,18 +282,256 @@ class CustomAssetsBuilder {
...
+  }
+
+  private u32pair(a: number, b: number): Uint8Array {
+    const data = new Uint8Array(8);
+    bufWriteU32BE(data, 0, a);
+    bufWriteU32BE(data, 4, b);
+    return data;
+  }
+
+  async addHumanAgeProperties() {
```
Nax, line 311: I don't fully understand what this does. Is it creating some sort of patch that gets applied at runtime to set child/adult properties?

packages/generator/lib/combo/custom/index.ts
```diff
@@ -476,17 +727,23 @@ class CustomAssetsBuilder {
...
+    await this.addAdultAgeModelTables(mmAdultLink);
+    await this.addHumanAgeProperties();
     await this.addObjectFile('MM_ADULT_EPONA', 'mm_adult_epona.zobj', []);
     await this.addObjectFile('MM_ADULT_LINK_SPIN_ATTACK_VTX_1', 'mm_adult_link_spin_attack_vtx_1.bin', []);
     await this.addObjectFile('MM_ADULT_LINK_SPIN_ATTACK_VTX_2', 'mm_adult_link_spin_attack_vtx_2.bin', []);
     await this.addObjectFile('MM_ADULT_LINK_SPIN_ATTACK_VTX_3', 'mm_adult_link_spin_attack_vtx_3.bin', []);
     await this.addObjectFile('MM_ADULT_LINK_MASK_MTX', 'mm_adult_link_mask_mtx.bin', []);
+    await this.addObjectFile('MASK_ADULT', 'object_gi_maskadult.zobj', [0x060009B0, 0x06000B90,]);
+    await this.addObjectFile('MASK_ADULT_TRANSFORM_PLAYER', 'object_mask_adult.zobj', [0x0a000900,]);
+    await this.addObjectFile('ADULT_MASK_EQUIPMENT', 'adult_mask_equipment_standalone.zobj', [0x0a000920,]);
```
Nax, line 746: Does this work safely with the code overrides we perform for custom equipment?

packages/logic/src/world/transform.ts
```diff
@@ -970,6 +970,10 @@ class LogicPassWorldTransform {
       this.addItem(Items.MM_MASK_SPOOKY);
     }
```
Nax, line 970: I think this file also defined the plentiful pool, should be included (you can splat the mask transform set, would be easier)

data/macros/macros_mm.yml
```diff
@@ -56,8 +56,8 @@
 "is_valley_cleared": "flag_on(MM_REGION_VALLEY_CLEARED) && flag_off(MM_REGION_VALLEY_CURSED) && ((setting(erBoss, none) && setting(regionState, dungeonBeaten) && event(BOSS_TWINMOLD)) || event(CLEAR_STATE_IKANA))"
 
 # Age
-"is_child": "!setting(crossAge) || age(child)"
-"is_adult": "setting(crossAge) && age(adult)"
+"is_child": "!setting(crossAge) || age(child) || has_mask_adult || setting(startingAgeMm, child)"
```
Nax: This will not work.
Two things:

1) With cross age the player can get into MM, use adult mask, then go back to OoT I think? This needs to be handled on the pathfinder level. Also, OoT doesn't like when you become adult without reaching temple of time first, need to make sure this works fine, especially with age swap in OoT.

2) With cross age the starting MM age needs to be completely ignored. The age will be whatever it was in OoT, OR altered by the adult mask (but see above)

Nax (commented):
Couple comments but otherwise pretty good!
The pathfinder/logic needs some work before this can be reliable.

packages/generator/lib/combo/custom/index.ts
Chronolegionnaire: This creates two serialized PlayerAgeProperties blobs rather than a list of runtime writes.

The child blob is copied directly from MM's existing Human properties. The adult blob starts from the Zora properties and overrides the values needed for Adult Link, then when swapping age, it grabs the appropriate blob and applies all the properties.

It was actually done this way instead of writing the properties directly in C because when I had initially implemented the swap, the mm payload had hit its limit so I had to find a way to minimize c code to reduce the pay load size. After I had gotten it working we upped the payload size so its not strictly required to be done this way anymore but it was already working so i didn't change it.

packages/generator/lib/combo/custom/index.ts
Chronolegionnaire: yep should be safe from my testing, these are separate objects and don't replace or mutate the custom-equipment objects themselves and the age switch reapplies the model tables before UpdateEquipment(), so the equipment overrides see the correct child/adult model pointers after the reload.

packages/logic/src/world/transform.ts
Chronolegionnaire: Fixed!

data/macros/macros_mm.yml
Chronolegionnaire: would something like this work for the pathfinder?

```private reExploreMmWithAdultMask(worldId: number) {
    if (!this.settings.crossAge)
      return;

    const world = this.worlds[worldId];
    const ws = this.state.ws[worldId];

    for (const age of AGES) {
      const otherAge = age === AGE_CHILD ? AGE_ADULT : AGE_CHILD;
      const areas = [...ws.ages[age].areas];
      for (const [area, areaData] of areas) {
        if (world.areas[area].game !== 'mm')
          continue;
        this.exploreArea(worldId, otherAge, area, cloneAreaData(areaData), area);
      }
    }
  }
  
  and call it at the end of requeue item with 
 if (item === Items.MM_MASK_ADULT) {
      this.reExploreMmWithAdultMask(worldId);
      
      
const canTimeTravel = ws.events.has(EVENT_TIME_TRAVEL) && worldArea.ageChange;
    const canUseAdultMask = this.settings.crossAge && worldArea.game === 'mm' && ws.items.has(Items.MM_MASK_ADULT);
    if ((canTimeTravel || canUseAdultMask) && area !== fromArea) {
    
    and this for the age swap

Chronolegionnaire: just gonna push it so its easier to see the code, lemme know if I need to add anything else

Chronolegionnaire: okay I'm getting a little stuck on how best to update the pathfinder. Let me know if you have any ideas or want to take a whack at it.

Chronolegionnaire: pushed some changes to the pathfinder, lemme know if I overlooked anything

Nax:
Closing to import into the main repo, see: #1033

PR #1030, opened 2026-08-24
Numerous Logic fixes and improvements regarding skulltulas + miscellaenous logic fixes

eedefeed:
I am very sorry how big this PR is. It just kept getting hideously complex.

I have not yet fully changed the changelog. I think I'll make a separate PR for that after merging.

Changes marked † in the change summary have details in the 'Change Detail' section

# Change Summary 

Consistently use the spelling "skulltula" in logic files. 
Numerous comments added to logic files, esp regarding alternate strats not accounted for yet (mostly within remit of this PR: skulltulas)

## New tricks/glitches
- Add Skulltula Elevator glitch (https://www.youtube.com/watch?v=s-66puPn368&list=PLOf1x45O7FsY)
- Trick for standing closely in front of the skulltula in forest MQ's first room (combines with skulltula elvator, hammer and bombchu as ways to clear the skulltula) (https://www.youtube.com/watch?v=zcr465LTwEU)

## macros_oot 
- Add temp_flag_event macro to handle situations where temp flags mean a single (ageless!) event is inappropriate
- † Add commented-out temp_flag_event_time_travel for temp_flag_event-type situations where time travel may not be available after moving to a new area
- Add is_forest_cleared macro
- Move icicle-related macros to a more appropriate section

- Add can_kill_skulltulas macro
- general changes to get_past_skulltulas_no_stone macro

## Kokiri Forest 
- logic now takes into account the skulltula that blocks adult link (mido tunnel to deku tree)

## Bottom of the Well 
- use get_past_skulltulas macro for initial skulltula
- add climb anywhere as a way past this skulltula

## Deku Tree 
- use get_past_skulltula macro for getting past the top-of-the-map skulltulas

## Forest Temple
- use get_past_skulltulas macro for first corridor's skulltula

## Thieves' Hideout
- fix logic's check for stone mask: `has_mask_stone` -> `can_use_mask_stone`. It now checks you can actually wear it as current age.

## Shadow Temple
- Scythe Silver Rupees - just some useful notes

## MQ Shadow
- Scythe Silver Rupees:
    - Account for the skulltulas that block the silver rupees
    - Add note about how vanilla SR collection logic flaw needs to be accounted for here
- Wind tunnels - overhaul getting past the skulltula and past the pit logic. Skulltula elevator won't work without a way to avoid getting carried away by wind, so we can't use the can_get_past_skulltulas macro.

## MQ Deku Tree 
- † The skulltula above the song of time block (basement) can be got around with nothing as both ages. Updated logic to reflect this.

## MQ Dodongo's Cavern 
- Consistently use "DC_MQ_" as the prefix for events rather than mixed use of "DC_MQ_" and "MQ_DC_"
- Staircase Room: properly account for the skulltulas by:
    - Splitting the upper area into two regions (Walkways, Upper Near Door), bounded by the skulltulas
    - Climb anywhere can be used to get around these skulltulas by climbing down onto the side of the walkway
    - split silver rupees between the two new regions as appropriate
- † Changes to the sequence of rooms: Vanilla Bomb Bag Room (lower) <-> before upper lizalfos <-> upper Lizalfos 
    - add additional region for the end of the corridor (just before the lizalfos fight)
    - overhaul events: remove some events that use temp flags. This avoids issues where logic expects the puzzle to still be solved if the player returns as a different age
    - locations/checks that relied on these events now check for the item requirements directly, even if that means crossing loading zones with lit deku sticks (which I assume door rando does not handle)
    - Add event, only triggerable when player has sot-age-change, that accounts for the ability to swap age (preserving the temp flag) and walk through the door to the larvae room (which may have age swap restricted in it e.g. Barinade boss room)
    - allow movement through invisible one-way collision (a web)
    - add event for destroying the mudwall at corridor end; this is a permanent flag

## MQ Forest Temple 
- First room skulltula overhauled.
    - add jump slash as a method (except with hammer)
    - remove spin slash as a method (jump slash covers this).
    - fixes raw bomb bag check - you need renewable bombs.
- † Movement from West garden to East garden via the high cooridor is blocked by 2 webs and a skulltula. The webs share a temp flag
    - split web destruction event into separate child and adult events.
    - use new temp_flag_event macro for this split
    - when climbing the vines, also check for the ability to either kill skulltula or use child+stone mask to get by it.
    - As a (probably too restrictive for door rando) failsafe, logic now understands that going through the door in the high alcove in the west garden now places link on the other side of the web, preventing him from going back through the door unless the web has been destroyed.
    - Sticks is now logical for destroying the web from the corridor, as long as you have access to the torch.
    - moving through the invisible one-way web now logical
    - moving through the visible one-way web is trick-gated (whereas previously it was merely web-destruction gated)
- Basement skulltula: logic now understands you can get past this with nothing

# Change Detail 

## MQ Deku Tree 
"The skulltula above the song of time block (basement) can be got around with nothing as both ages. Updated logic to reflect this."

Whilst the positioning to do this as child is precise, that positioning is both intuitive and mechcanially easy. You walk into the corner between the block and the wall (either side) and that makes sure you are as close to the wall as possible. Then climb the ledge without further ground movement else you might get knocked off. I can see a case for it being a trick, but I think that if something is intuitive and easy, it shouldn't be a trick.

As adult, it's much easier.

## MQ Dodongo 

MQ Dodongo: aside from adding new regions (neccesary because of the skultulas), I had to change how events worked in the upper east area (vanilla bomb bag room through to the lizalfos fight). This is because most of these events are temp flags. Because OoTMM's implementation of events is that they are collected by a single age and then usable by both ages, this meant logic was too permissive - in rare cases it would be possible for the generator to want you to trigger the event as one age and then come back as a different age to use that event, even though that's not possible. (Josh uses a lot of events in his MQ reworks, I reckon there's quite a few of these issues lurking).

An additional wrinkle was Song-of-time-age-change, which does not reset the temp flags when used. Without age change, you can only enter the _larvea room door_ if you can beat the 3-torch puzzle as that age (since if you leave the temple to change age, the door gets rebarred). If you have access to age change, you can enter the door as any age. This was implemented as an event (`DC_MQ_SOLVE_3-TORCH_PUZZLE_TIME_TRAVEL`) to share success in one age with the other.

The flags in this part of DC are as follows:
```
Things activated by 3-torch puzzle:
    the gold torch in that room                         --- TEMP, survives age change
    the door to larvae                                  --- TEMP, survives age change
the web torches act as one:
    if only one is lit, other goes out
    when both are lit, the gold torch lights
    survives room reload but not dungeon reload
    These and the gold torch they light are all:        --- TEMP, survives age change
the web                                                 --- PERM
the mud wall                                            --- PERM
```

## MQ Forest Temple 
The high web in forest temple MQ's East Courtyard (guarded by a skulltula) is also a temp flag. This flag survives age change. We have to solve this with a different implementation to Dodongo MQ because:
- Destroying one of the webs destroys the other web too (in a different room)
- There are multiple places (in logic) from which you can destroy the web
- I don't think there's a need to specially account for sot-age-change because either side of both webs is guaranteed a place you can use SoT age change.

So the implementation I chose here was to make the macro that checks event ages and double-up each event as a child/adult version.

Door rando may need to reconsider how these webs' shared flags work.

## temp_flag_event_time_travel macro 
I don't know if this macro is necessary. In Dodongo MQ I solved the temp flag + age change problem much more elegantly than this macro. Perhaps there are use-cases for the macro that I can't think of. At any rate, it's not used and therefore is commented out.

# Additional info 

Some skulltulas are too low for adult link to walk under using stone mask. But adult link can still roll under. Base logic now expects the player to know this.

I haven't touched logic for the skulltulas in deku basement (vanilla layout). You can get by both with nothing, but the one in water/spike room is a little awkward and not intuitive. The MQ one is pretty intuitive to get around.

There's a strange interaction with stone mask and skulltulas that block passageways. Running into them with stone mask often results in Link being shoved in a _random_ direction. This effectively means stone mask is enough to get past all skulltulas. Is this a glitch that should be patched?

There are many skulltulas that you can get past with nothing by getting a bit lucky and getting knocked past them. The only one I've made logical is forest temple MQ, because it's easy and semi-reliable. I think it could be a general purpose trick.

I think there needs to be a broader look at the interaction of events and temp flags, esp in Joshua's commits, since his style uses a lot of events. (And MQ seems to use a lot of temp flags)

Thanks for coming to my _series_ of TED talks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ All notable changes to this project will be documented in this file.
 - Fix moon gossip stones sometimes hinting non-vanilla masks when item extensions are enabled.
 - Fix logic for most of the pots in Great Bay Coast not checking for the ability to swim.
 - Fix logic for GBC Tingle's photo not checking for the ability to knock him down.
+- Fix logic not understanding that the Kokiri Forest (adult) Skulltula can block the player
 
 ## [32.0] - 2026-07-19
 
--- a/data/macros/macros_oot.yml
+++ b/data/macros/macros_oot.yml
@@ -227,8 +227,8 @@
 "can_hit_scrub": "has_nuts || can_hit_triggers_distance || has_shield_for_scrubs || can_collect_distance || can_hammer || has_explosives"
 "can_rescue_carpenter": "small_keys_hideout_all && (can_use_sword || ((can_boomerang || has_nuts) && can_use_sticks)) && soul_carpenters && soul_thief_fighter"
 "carpenters_rescued": "setting(gerudoFortress, open) || (event(CARPENTER_1) && (setting(gerudoFortress, single) || (event(CARPENTER_2) && event(CARPENTER_3) && event(CARPENTER_4))))"
-"get_past_skultullas": "get_past_skultullas_no_stone || can_use_mask_stone"
-"get_past_skultullas_no_stone": "has_nuts || can_use_sword_or_sticks || has_ranged_weapon || can_use_din || has_explosives || can_hammer || can_use_mask_blast || trick(GLITCH_OOT_SKULLTULA_ELEVATOR)"
+"get_past_skulltulas": "get_past_skulltulas_no_stone || can_use_mask_stone"
+"get_past_skulltulas_no_stone": "has_nuts || can_use_sword_or_sticks || has_ranged_weapon || can_use_din || has_explosives || can_hammer || can_use_mask_blast || trick(GLITCH_OOT_SKULLTULA_ELEVATOR)"
 "collect_gs_on_walls": "gs && (can_collect_distance || (has_ranged_weapon && climb_anywhere))" #TODO: Low/high macros like beehives; Din's, Blast mask, bomchus, bombs (if within throwing distance) and maybe even jump slashes if space permits (and damage?)
 "can_carry_keg": "can_lift_gold"
 "can_use_keg": "can_carry_keg && cond(setting(sharedPowderKeg), event(MM_BUY_KEG) || event(BUY_KEG), setting(powderKegOot) && event(BUY_KEG))"
@@ -243,7 +243,7 @@
 # Mido / Deku Tree
 "can_move_mido_reqs": "is_child && has_sword_kokiri && has_shield_deku"
 "can_move_mido": "can_move_mido_reqs && soul_npc(SOUL_NPC_MIDO)"
-"can_bypass_mido": "setting(dekuTree, open) || climb_anywhere || hookshot_anywhere || can_move_mido_reqs || is_adult && (get_past_skultullas_no_stone || is_forest_cleared)"
+"can_bypass_mido": "setting(dekuTree, open) || climb_anywhere || hookshot_anywhere || can_move_mido_reqs || is_adult && (get_past_skulltulas_no_stone || is_forest_cleared)"
 
 #Specialty Macros
 "met_zelda": "event(MEET_ZELDA) || setting(skipZelda)"
--- a/data/macros/macros_oot.yml
+++ b/data/macros/macros_oot.yml
@@ -1,6 +1,8 @@
 #Generic Utility Macros
 "is_child": "age(child)"
 "is_adult": "age(adult)"
+"temp_flag_event(childEvent, adultEvent)": "(is_child && childEvent) || (is_adult && adultEvent)"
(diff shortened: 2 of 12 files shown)


eedefeed:
There is another change I forgot to document:

Shadow MQ wind tunnel: "hookshot_anywhere" was mentioned as a way to get through the wind tunnel. Just hookshot is enough in either direction.

PR #1037, opened 2026-08-29
Alphabetize and make shared names consistent.

Chronolegionnaire:
Alphabetize and make shared names consistent.


Nax:
Merged manually after a rebase+sort 👍

PR #1039, opened 2026-08-30
Items tab reorg

Chronolegionnaire:
Reorganize items extensions tab and fold over shared and progressive items into this tab as well


Nax:
Had to squash because the merge created phantom conflicts, but was merged 👍

PR #1045, opened 2026-09-04
Bottle tweaks

Chronolegionnaire:
Various tweaks and fixes related to bottles:

Add a setting to add extra OoT bottles without having to share bottles.

When more than 4 OoT bottles are found (when pool is set to plentiful), you no longer gain extra trade slot bottle slots unless Extra OoT Bottles is on.

When random bottle contents is on, random bottles can be added to plando or starting items.

Plentiful item pool now properly has an extra shared bottle when bottles are shared.

Fix bottles not having a consistent amount in pool under certain combinations of settings. (zora king and shared bottles)

Fix equip swapped items from reverting to bottles on game change if they were equip swapped over bottles.


packages/core/src/settings/data.ts
```diff
@@ -2375,6 +2375,14 @@ export const SETTINGS = [{
   description: "Add the Powder Keg in Ocarina of Time. Once found, Medigoron in Goron City will sell Powder Kegs after you buy his item.",
   default: false,
   cond: hasOoT,
+}, {
+  key: 'extraBottlesOot',
+  name: 'Extra Bottles (OoT)',
+  category: 'items.extensions',
```
Nax: Does this not need to be adapted to the item refactor PR?

Chronolegionnaire: Updated to new format

Nax (approved):

PR #1048, opened 2026-09-10
Dedupe

Chronolegionnaire:
Deduplicate sound and player animation data to recover more rom space


Chronolegionnaire:
Closing cause I found some issues with this implementation around custom music

PR #1049, opened 2026-09-13
Adult link mm voices are now a separate pool entirely 

Chronolegionnaire:
Adult link mm voices are now a separate pool entirely from fierce deity and grow the sound bank safely so background effects are not destroyed.

Fixes background effects on moon/ikana graveyard/etc


Nax:
Seems like this also includes unrelated models/cosmetics stuff?

Chronolegionnaire:
> Seems like this also includes unrelated models/cosmetics stuff?

Yep, I made this from modifying the custom cosmetics stuff I was working on. Should be good now

packages/generator/src/mm/actors/Player.c
```diff
@@ -1745,6 +1779,25 @@ void Player_SkelAnime_DrawFlexLod(PlayState* play, void** skeleton, Vec3s* joint
         postLimbDraw = Player_PostLimbDrawGameplayWrapper;
 
     sPlayerOverrideLimb = overrideLimbDraw;
+    if (player->transformation == MM_PLAYER_FORM_HUMAN && comboIsLinkAdult())
```
Nax: Isn't that also unrelated?

Chronolegionnaire: Fixed...again

PR #1050, opened 2026-09-14
MM age requirements rebased

Chronolegionnaire:
MM age requirements rebased


packages/generator/src/common/item/item_add.c
```diff
@@ -822,9 +820,9 @@ static void addHookshotRawMm(PlayState* play, int level)
         itemId = ITEM_MM_HOOKSHOT;
     else
         itemId = 0x11; /* ITEM_MM_BOTTLE_POTION_RED but that enum is wrong */
-    gMmSave.info.inventory.items[ITS_MM_HOOKSHOT] = itemId;
+    if (gMmSave.info.inventory.items[ITS_MM_HOOKSHOT] == ITEM_NONE)
+        gMmSave.info.inventory.items[ITS_MM_HOOKSHOT] = itemId;
     gMmExtraItems.hookshot |= (1 << (level - 1));
-    reloadSlotMm(play, ITS_MM_HOOKSHOT);
```
Nax, line 827: Why don't we reload the slot here? If we got the longshot we should make it replace the hookshot by default.

packages/generator/src/common/item/item_add.c
```diff
@@ -948,9 +946,9 @@ static void addOcarinaRawMm(PlayState* play, int level)
         itemId = ITEM_MM_OCARINA_OF_TIME;
     else
         itemId = ITEM_MM_OCARINA_FAIRY;
-    gMmSave.info.inventory.items[ITS_MM_OCARINA] = itemId;
+    if (gMmSave.info.inventory.items[ITS_MM_OCARINA] == ITEM_NONE)
+        gMmSave.info.inventory.items[ITS_MM_OCARINA] = itemId;
     gMmExtraItems.ocarina |= (1 << (level - 1));
-    reloadSlotMm(play, ITS_MM_OCARINA);
```
Nax, line 953: Same with the ocarina

packages/generator/src/mm/trade.c
```diff
@@ -16,6 +16,28 @@ static void removeButtonItem(u16 itemId)
     }
 }
 
+static void removeCurrentCButtonsForSlot(u16 slot)
```
Nax: I think there was a duplicate of this function in another file

Nax (approved):
Left a few minor comments, mostly okay.

Not a fan of the way the settings are encoded (with a pair of booleans) - it creates an illegal state (both false). Seems like it would be easier to have them as 'child', 'adult', 'both' enums instead?

But this can be tackled in an eventual v2.

Good job otherwise.

Nax:
Also the confvar logic is a bit weird and confused me at first - it took me a moment to realize both convars being false means either age, meaning the flags really mean "GATE at this age" instead of "PERMIT usage at this age".

packages/generator/src/common/item/item_add.c
Chronolegionnaire: I got rid of the reload here because it would result in a situation where you have the wrong age locked item equipped if for example you had short hookshot equipped (and locked to child) and then picked up another hookshot upgrading it to longshot which can be locked to the opposite age. Same thing with the ocarinas

Chronolegionnaire:
Don't know why I didn't think to just do a bunch of enums, definitely would have been cleaner that way. Will put on my list to rework that.

packages/generator/src/mm/trade.c
Chronolegionnaire: Fixed!

PR #1051, opened 2026-09-14
add .aseq and .seq to mmrs sequence file search regex

zl64:
`.mmrs` format parity with MMR per ZoeyZolotova/mm-rando@9ea3eb2


Nax:
Thanks for the heads up, used `(z|a)?seq` as a slightly more readable regex.

PR #1052, opened 2026-09-14
fix: Fix GBT Map Room Bronze Scale Logic Issue

Joshua8600:
Logic incorrectly though you can use just hookshot to get from the GBT Map Room floor to the Map Chest, ignoring the fact that you need to surface.

This PR aims to resolve that and associated issues in the Baba Room as well by adding an "Underwater" region to both rooms.

Discussion on Discord starts here: <https://discord.com/channels/1004394204992118935/1045066945512345692/1549090259256741949>


data/world/mm/dungeons/great_bay_temple.yml
```diff
@@ -49,7 +49,7 @@
   dungeon: GB
   exits:
     "Great Bay Temple Water Wheel": "true"
-    "Great Bay Temple Map Room": "underwater_walking || hookshot_anywhere || (trick(MM_GBT_CENTRAL_GEYSER) && ((can_use_fire_arrows && can_use_ice_arrows) || can_use_din || can_use_farore || can_use_nayru))"
+    "Great Bay Temple Map Room Underwater": "underwater_walking || hookshot_anywhere || (trick(MM_GBT_CENTRAL_GEYSER) && ((can_use_fire_arrows && can_use_ice_arrows) || can_use_din || can_use_farore || can_use_nayru))" #Does Underwater Walking here need strict for the time?
```
XenoWars:
```suggestion
    "Great Bay Temple Map Room Underwater": "underwater_walking || (can_swim && (hookshot_anywhere || (trick(MM_GBT_CENTRAL_GEYSER) && ((can_use_fire_arrows && can_use_ice_arrows) || can_use_din || can_use_farore || can_use_nayru))))"
```
The timer isn't anything to worry about. 24s is way more than enough to make it to the room itself. I also tested collecting all the pots as well and you can collect all of them within the same 24s with a bit of room for error.

data/world/mm/dungeons/great_bay_temple.yml
```diff
@@ -87,20 +81,47 @@
...
+    "Great Bay Temple Map": "has_mask_zora || short_hook_anywhere || can_use_ice_arrows || can_hookshot || has(MASK_DEKU)"
+    "Great Bay Temple SF Map Room Pot": "has_mask_zora || short_hook_anywhere || can_use_ice_arrows || has(MASK_DEKU)"
+    "Great Bay Temple Pot Map Room Surface 1": "true"
+    "Great Bay Temple Pot Map Room Surface 2": "true"
+    "Great Bay Temple Pot Map Room Surface 3": "has_mask_zora || short_hook_anywhere || can_use_ice_arrows || has(MASK_DEKU)"
+
+"Great Bay Temple Baba Room Underwater":
+  dungeon: GB
+  events:
+    GBT_BABA_CHEST_SPAWN: "soul_enemy(SOUL_ENEMY_BIO_BABA) && (has_mask_zora || (underwater_walking && has_arrows))"
```
XenoWars: Can't use arrows underwater.

data/world/mm/dungeons/great_bay_temple.yml
```diff
@@ -63,21 +63,15 @@
...
+"Great Bay Temple Map Room Underwater":
   dungeon: GB
   events:
     MAGIC: "true"
   exits:
-    "Great Bay Temple Baba Room": "has_mask_zora || (((has_bombchu && trick(MM_GBT_BABA_ENTRY_BOMBCHU)) || has_mask_blast || has_boomerang) && (short_hook_anywhere || underwater_walking || (can_dive_big && has_tunic_zora))) || hookshot_anywhere"
-    "Great Bay Temple Red Pipe 2": "can_use_ice_arrows || short_hook_anywhere"
-    "Great Bay Temple Green Pipe 3": "short_hook_anywhere"
+    "Great Bay Temple Map Room Surface": "can_swim || has_mask_zora || short_hook_anywhere" #Might need long hook anywhere, needs testing
+    "Great Bay Temple Central Room": "underwater_walking || hookshot_anywhere" #Underwater walking only leads to bottom of the room, so not entirely accurate. Separate bottom and surface in dungeon rework.
```
XenoWars: Can't go backwards through here unless the water is going the other way. You would also need short_hook anywhere with iron boots.

data/world/mm/dungeons/great_bay_temple.yml
```diff
@@ -63,21 +63,15 @@
...
-"Great Bay Temple Map Room":
+"Great Bay Temple Map Room Underwater":
   dungeon: GB
   events:
     MAGIC: "true"
   exits:
-    "Great Bay Temple Baba Room": "has_mask_zora || (((has_bombchu && trick(MM_GBT_BABA_ENTRY_BOMBCHU)) || has_mask_blast || has_boomerang) && (short_hook_anywhere || underwater_walking || (can_dive_big && has_tunic_zora))) || hookshot_anywhere"
-    "Great Bay Temple Red Pipe 2": "can_use_ice_arrows || short_hook_anywhere"
-    "Great Bay Temple Green Pipe 3": "short_hook_anywhere"
+    "Great Bay Temple Map Room Surface": "can_swim || has_mask_zora || short_hook_anywhere" #Might need long hook anywhere, needs testing
```
XenoWars: You can't stand in the tunnel, so this will hard require hookshot over short hook.

data/world/mm/dungeons/great_bay_temple.yml
```diff
@@ -87,20 +81,47 @@
     "Great Bay Temple Pot Map Room Water 7": "underwater_walking || hookshot_anywhere"
     "Great Bay Temple Pot Map Room Water 8": "underwater_walking || hookshot_anywhere"
```
XenoWars: Hookshot anywhere shouldn't be in the underwater logic for this since to be underwater and able to use hookshot you need to be able to walk underwater in the first place. Can add it as logic to the surface region though.

data/world/mm/dungeons/great_bay_temple.yml
```diff
@@ -87,20 +81,47 @@
...
+  events:
+    GBT_BABA_CHEST_SPAWN: "soul_enemy(SOUL_ENEMY_BIO_BABA) && (has_mask_zora || (underwater_walking && has_arrows))"
+  exits:
+    "Great Bay Temple Baba Room Surface": "has_mask_zora || can_swim || (underwater_walking && short_hook_anywhere)"
+    "WARP_SONGS": "true"
+
+"Great Bay Temple Baba Room Surface":
+  dungeon: GB
+  events:
+    GBT_BABA_CHEST_SPAWN: "soul_enemy(SOUL_ENEMY_BIO_BABA) && (has_mask_zora || has_arrows || (has_mask_blast && short_hook_anywhere))"
```
XenoWars: Hookshot can work on the underwater babas, and can knock down the hanging one to then damage it with sword or other item that would defeat it. don't need to have anywhere on.

XenoWars (changes requested):

data/world/mm/dungeons/great_bay_temple.yml
Joshua8600: Changed it to hookshot or boomerang

data/world/mm/dungeons/great_bay_temple.yml
Joshua8600: Fixed, but commented out line in general. Not deleting in case we need logic for reversed water flow.

data/world/mm/dungeons/great_bay_temple.yml
Joshua8600: Fixed.

data/world/mm/dungeons/great_bay_temple.yml
Joshua8600: Fixed and added

data/world/mm/dungeons/great_bay_temple.yml
Joshua8600: Fixed

data/world/mm/dungeons/great_bay_temple.yml
XenoWars: also can't use boomerang underwater, and hookshot doesn't defeat the hanging one even if it goes underwater. Really, the event can just be axed from this region. The only possible non-zora method would be hook+blast+irons+tunic (24s is not enough), but to even be here without the ability to swim you need hook anywhere which just gets you to the surface anyway. And if you have zora, you can surface anyway and trigger the event there.

Joshua8600: > Really, the event can just be axed from this region.

Yeah, I just removed it,

PR #1056, opened 2026-09-20
Fix restrictive Powder Keg Trial logic

XenoWars:
Fixes the Powder Keg Trial being too restrictive, forcing goron outside ER unless either of the two tricks for golden gauntlets were enabled.


XenoWars:
Revven said it was ok in discord.

